PR #99, opened 2022-08-09
Add `tox` as a [dev] dependency

maarten-ic:
When following the instructions on https://muscle3.readthedocs.io/en/latest/contributing.html, the `make test` command fails with the error:

```
[...]
make: tox: Command not found
make: *** [Makefile:30: test_python] Error 127
```

This PR adds tox as a development dependency to avoid this error.


LourensVeen:
I have tox installed globally, so I guess that's why I didn't miss it. And the CI install tox explicitly. Maybe it would actually be better to replace that by a call to installing the [dev] dependencies? Then we can actually test that it works.

PR #100, opened 2022-08-09
`make test` issues

maarten-ic:
This PR fixes two issues:
- Flake8 check (which was probably added in the 5.0.x release of flake8 from end of July) failed with `libmuscle/python/libmuscle/instance.py:443:20: E275 missing whitespace after keyword`
- `integration_test/test_mpi_macro_micro.py` did not explicitly provide connection details to the `macro` instance. This works as long as `muscle_manager` is able to bind to the default port 9000, but failed when another program has claimed that port.


maarten-ic:
Replaced by #107

PR #108, opened 2022-08-26
Implement checkpoint triggers and tests

maarten-ic:


tox.ini
```diff
@@ -8,6 +8,7 @@ deps =
     flake8
     pytest
     pytest-cov
+    git+https://github.com/multiscale/ymmsl-python.git@feature/checkpointing#egg=ymmsl
```
LourensVeen, line 11: Does this work with the latest pip? I had issues before because they removed this functionality in recent versions, but there was a work-around, IIRC some way of having GitHub make a tarball on demand which pip can then install.

LourensVeen: I looked into this some more, and it seems git+git has been removed, but git+https should still work. So I'm confused now as to what problems I had. If it works, it works I guess.

LourensVeen (approved):
Nice! I added a comment about pip support for installing from GitHub, which I had problems with before on recent versions of pip. I can't find anything back about it though, so if this works then it works.

PR #111, opened 2022-09-05
Feature/checkpointing instance

maarten-ic:
Builds on top of #110


maarten-ic:
Complete implementation is part of #116, closing this PR.

PR #114, opened 2022-09-15
Reduce delays

maarten-ic:
- instance_manager.py:
  - Increase polling frequency of LogHandlingThread. Reduces the shutdown delay (in InstanceManager.shutdown), saves 0-900 ms.
- qcgpj_instantiator.py
  - Add a 10ms delay at the start of QCGPJInstantiator._main to allow the main process some time for submitting InstantiationRequests. Saves 90ms in startup duration.
  - Do not sleep in QCGPJInstantiator._main when a shutdown request is received. If all instances exited successfully we can immediately be done and save a 100ms wait.
- tcp_transport_server.py
  - Set poll_interval of SocketServer.serve_forever to 100ms. Saves 0-400 ms in shutdown duration (in instances and muscle_manager).

Combined, these changes save 190ms to ~2 seconds for a run started with `muscle_manager	--start-all`. This is most notable for short runs, like the ones in the unit tests.

`make test` duration, averaged over 5 runs (no compilation):
- Before this commit: 50.51 seconds
- After this commit: 29.51 seconds

### DIFF
--- a/libmuscle/python/libmuscle/manager/qcgpj_instantiator.py
+++ b/libmuscle/python/libmuscle/manager/qcgpj_instantiator.py
@@ -153,7 +153,6 @@ class QCGPJInstantiator(mp.Process):
         shutting_down = False
         done = False
         while not done:
-            do_sleep = True
             while not shutting_down:
                 try:
                     request = self._requests_in.get_nowait()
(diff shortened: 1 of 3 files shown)


libmuscle/python/libmuscle/manager/qcgpj_instantiator.py
```diff
@@ -178,7 +182,8 @@ async def _main(self) -> None:
                 except queue.Empty:
                     break
 
-            await asyncio.sleep(0.1)
```
LourensVeen, line 181: This code is a bit complicated, and I needed to read it again, but it seems to me that we can just move the sleep statement into an `if not done:` block at the end of the outer loop? That way we first drain the requests queue and submit all the launch and cancel requests to QCG-PJ, and then we process all finished processes right away, and then we check whether we should quit and if not, sleep. And we avoid an extra variable. What do you think?

PR #119, opened 2022-10-18
Ensure correct timestamp type in Message

### LINKED ISSUE #118: Passing a non-float for the Message timestamps in Python results in invalid message on the wire
The API requires Message timestamps to be floats, but of course Python doesn't type check and neither does libmuscle, so if you pass an int it'll get serialised to MsgPack, resulting in a message on the wire that has the wrong format.

In communicator.py, the Message constructor should try to convert the value to float, just in case.

Thanks to Jonathan for finding this one!

maarten-ic:
Fixes #118 

When an incorrect type is provided by the user for Message.timestamp or Message.next_timestamp, MsgPack will serialize an invalid MMPMessage on the wire. This leads to errors in statically typed peer actors.

Issue is fixed by explicitly converting to float in Message.__init__ and checking again in MMPMessage.__init__ (as the user may have assigned another value between creation of the Message and Instance.send).

### DIFF
--- a/libmuscle/python/libmuscle/communicator.py
+++ b/libmuscle/python/libmuscle/communicator.py
@@ -49,6 +49,11 @@ class Message:
             data: An object to send or that was received.
             settings: Overlay settings to send or that were received.
         """
+        # make sure timestamp and next_timestamp are floats
+        timestamp = float(timestamp)
+        if next_timestamp is not None:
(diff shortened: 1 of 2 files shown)


LourensVeen:
Looks good. Ideally people would use mypy or an IDE that does so for them, but if they don't and pass something that cleanly converts to a float, why not?

PR #127, opened 2022-11-07
Add LD_LIBRARY_PATH separators (`:`) in yMMSL implementation sections

### LINKED ISSUE #126: Integration test `test_start_all` fails when LD_LIBRARY_PATH is defined
# Pytest error:

```text
integration_test/test_start_all.py:65: AssertionError
-------------------------------------------------- Captured log call ---------------------------------------------------
INFO     libmuscle.manager.instance_manager:instance_manager.py:99 Planned macro on Resources(<redacted>: 0)
INFO     libmuscle.manager.instance_manager:instance_manager.py:99 Planned micro on Resources(<redacted>: 0)
INFO     libmuscle.manager.instance_manager:instance_manager.py:122 Instantiating macro on Resources(<redacted>: 0)
INFO     libmuscle.manager.instance_manager:instance_manager.py:122 Instantiating micro on Resources(<redacted>: 0)
ERROR    libmuscle.manager.instance_manager:instance_manager.py:144 Instance macro quit with error 127
ERROR    libmuscle.manager.instance_manager:instance_manager.py:147 Output may be found in /tmp/pytest-of-<user>/pytest-3/test_start_all0/run/instances/macro
ERROR    libmuscle.manager.instance_manager:instance_manager.py:144 Instance micro quit with error 127
ERROR    libmuscle.manager.instance_manager:instance_manager.py:147 Output may be found in /tmp/pytest-of-<user>/pytest-3/test_start_all0/run/instances/micro
```

Inspecting stderr.txt of either component gives:
```
<redacted>/muscle3/libmuscle/cpp/build/libmuscle/tests/component_test: error while loading shared libraries: libmsgpackc.so.2: cannot open shared object file: No such file or directory
```

# Root cause

The yMMSL configuration (taken from run directory) has the following implementations section:

```yaml
implementations:
  component:
    env:
      +LD_LIBRARY_PATH: <redacted>/muscle3/libmuscle/cpp/build/msgpack/msgpack/lib
    executable: <redacted>/muscle3/libmuscle/cpp/build/libmuscle/tests/component_test
```

Note that this adds the string `<redacted>/muscle3/libmuscle/cpp/build/msgpack/msgpack/lib` to the LD_LIBRARY_PATH env variable. However, when LD_LIBRARY_PATH is something (e.g `/path/A:/path/B`) when starting, the last two paths are concatenated without the `:`-separator. This results in the file-not-found error observed above.

maarten-ic:
See #126


maarten-ic:
Note: documentation already mentions that the colon needs to be added, so no need to update any docs.
https://muscle3.readthedocs.io/en/latest/cplusplus.html#describing-c-implementations

LourensVeen:
Ah, nice catch. I'm wondering now if there's some way of automating a test for this. Or maybe just reduce the duplication between the integration tests a bit. On the other hand, it's not a million copies either, and this diff doesn't look too frightening, so maybe we just leave it like this, we can always make an issue for when a sufficiently round tuit arrives.

PR #134, opened 2022-11-16
More unified integration tests with native code

maarten-ic:
@LourensVeen I've tried to unify running integration tests with combined python actors (as methods) and native actors. Please have a look at this and let me know if you agree with this approach.

Thanks!


LourensVeen (approved):
You're a hoopy frood is all I can say about this. Beautiful work!

PR #135, opened 2022-11-21
Set TCP options in Python

maarten-ic:
Set TCP_NODELAY and TCP_QUICKACK (like the C++ code already did)


LourensVeen (approved):
Ah, that hasattr() looked really fishy at first, but it's how it works apparently. The attribute only exists if the platform supports it. If that's what Python wants, that's what it gets :-).

PR #152, opened 2023-01-12
Feature/instance flags

maarten-ic:


libmuscle/python/libmuscle/instance.py
```diff
@@ -29,6 +31,62 @@
...
+    will tell you about it in an error message and you can add it still.
+    """
+
+    USES_CHECKPOINT_API = auto()
+    """Indicate that this instance supports checkpointing.
+
+    You may not use any checkpointing API calls when this flag is not supplied.
+    """
+
+    HAS_NO_STATE_FOR_NEXT_USE = auto()
```
LourensVeen, line 67: KEEPS_NO_STATE_FOR_NEXT_USE? I think it may have a state because it can still iterate, but there may not be any gain to reusing the previous one, or it may even be worse. Naming these is tricky though...

libmuscle/python/libmuscle/instance.py
```diff
@@ -29,6 +31,62 @@
...
+    reuse loop.
+
+    This corresponds to :external:py:attr:`ymmsl.KeepsStateForNextUse.NO`.
+
+    If neither :attr:`HAS_NO_STATE_FOR_NEXT_USE` and
+    :attr:`STATE_FOR_NEXT_USE_NOT_REQUIRED` are supplied, this corresponds to
+    :external:py:attr:`ymmsl.KeepsStateForNextUse.REQUIRED`.
+    """
+
+    STATE_FOR_NEXT_USE_NOT_REQUIRED = auto()
```
LourensVeen, line 78: STATE_NOT_REQUIRED_FOR_NEXT_USE? This reads a bit weird to me.

libmuscle/python/libmuscle/instance.py
```diff
@@ -126,7 +183,7 @@ def __init__(
         self._set_local_log_level()
         self._set_remote_log_level()
 
-    def reuse_instance(self, apply_overlay: bool = True) -> bool:
+    def reuse_instance(self, apply_overlay: Optional[bool] = None) -> bool:
```
LourensVeen, line 186: Given the API change, should we check this for compatibility with the flags? Or warn if it's specified that it will be ignored? I doubt anyone is using it, so I think it's okay if it doesn't work. Actually, then we could just remove it.

libmuscle/python/libmuscle/instance.py
```diff
@@ -965,12 +1015,20 @@ def __receive_settings(self) -> bool:
...
         return True
 
-    def __pre_receive_f_init(self, apply_overlay: bool) -> None:
+    def __pre_receive_f_init(self, apply_overlay: Optional[bool]) -> None:
         """Receives on all ports connected to F_INIT.
 
         This receives all incoming messages on F_INIT and stores them
         in self._f_init_cache.
         """
+        if apply_overlay is not None:
```
LourensVeen, line 1024: Oh hey, look at that. Never mind, this works :). We'll remove it at some point in the future.

LourensVeen (commented):
A few comments, you're triggering my language perfectionist tendencies again :). I'm pretty sure no one is using apply_overlay, but this solution is also fine with me, so I'll leave it to you.

PR #154, opened 2023-01-12
Implement version check

### LINKED ISSUE #148: Add version check when connecting to manager
MUSCLE3 does not support mixing library versions in one workflow. For example, when using the `muscle_manager` from the latest release (0.5.0) and an actor that uses the `feature/checkpointing` branch - which should become part of the next release - an obscure error is triggered:

```text
----------------------------------------
Exception happened during processing of request from ('127.0.0.1', 48936)
  File "/usr/lib/python3.8/socketserver.py", line 683, in process_request_thread
    self.finish_request(request, client_address)
  File "/usr/lib/python3.8/socketserver.py", line 360, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib/python3.8/socketserver.py", line 747, in __init__
    self.handle()
  File ".../venv/lib/python3.8/site-packages/libmuscle/mcp/tcp_transport_server.py", line 37, in handle
    response = server._handler.handle_request(request)
  File ".../venv/lib/python3.8/site-packages/libmuscle/manager/mmp_server.py", line 84, in handle_request
    return cast(bytes, msgpack.packb(response, use_bin_type=True))
UnboundLocalError: local variable 'response' referenced before assignment
----------------------------------------
```

The error is triggered by a new request type that did not exist in 0.5.0, but users will not easily know that.

**Desired behaviour**

`muscle_manager` should check the library versions of all connecting instances. If any instance is using a different library version than the manager has, a clear error should be displayed to notify users about the version mismatch.

maarten-ic:
Fixes #148


LourensVeen (approved):
Looks good! I think the additional dependency in the Makefile is because the compiler doesn't add version.h to the .d file because it doesn't exist when it's scanned? Would it be enough to only add it to the dependencies for the .d file?

Issue #125, opened 2022-11-06
Simplify Message creation syntax


LourensVeen:
In Python, to create a Message with no `next_timestamp`, you have to write `msg = Message(t_cur, None, data)`. It would be nice (especially for teaching purposes) if you could just skip that second argument, like you can in C++ and Fortran. You can then write `msg = Message(t_cur, data=data)` and not worry about next timestamps.

This would require setting a default value for `next_timestamp`, as well as for `data`. We could actually default both to `None` without issue. `Message(t_cur)` could even be nice idiom for an empty event.

PR #155, opened 2023-01-13
Improve Message construction syntax

### LINKED ISSUE #146: Improve Message construction syntax
In many cases, the `t_next` argument to `Message()` isn't used. In C++, there's a constructor that only takes `timestamp` and `data`, but in Python you have to explicitly specify `None` for t_next if you don't want to use it. It would be nice if you could omit it, using `Message(t_cur, data=data)`. For that, both `t_next` and `data` need to have a default value (data because of Python syntax requirements).

If we default both to `None`, then `Message(t_cur)` is also valid, and sends a nil value. That sounds reasonable enough, and may even be a nice syntax for sending an empty trigger message. We should be consistent and allow the same syntax in C++ then, with an extra `Message(double timestamp)` constructor.

maarten-ic:
Fixes #146, #125

- Python: add default (None) values for next_timestamp and data in Message.__init__
- C++: add Message(timestamp) constructor
- Fortran: add LIBMUSCLE_Message_create(timestamp)

Update Python examples to use `Message(t, data=...)` instead of `Message(t, None, ...)`

### DIFF
--- a/libmuscle/cpp/src/libmuscle/message.hpp
+++ b/libmuscle/cpp/src/libmuscle/message.hpp
@@ -21,7 +21,7 @@ class Message {
          *
          * @param timestamp Simulation time for which this data is valid.
          */
-        Message(double timestamp);
+        explicit Message(double timestamp);
 
         /** Create a Message.
          *


maarten-ic:
@LourensVeen Please decide if we should do something with the C++ warning: `Class 'Message' has a constructor with 1 argument that is not explicit.`

LourensVeen:
Hmm. `instance.send(10.0)` looks like we're sending the value `10.0`, but it would actually send an empty Message with a timestamp at `10.0`. That's really unintuitive. If we make it explicit, people would have to write `instance.send(Message(10.0))`, which to be honest doesn't look much better. On the other hand, C++ doesn't have named arguments, so there's not much we can do about that.

Ah, but in a real model, it would actually read `instance.send(t_cur)` or something, which would make it clearer that it's a timestamp. So it's between that and `instance.send(Message(t_cur))`.

Kind of a toss-up, but I think explicit is better than implicit, so the BDFL is going to pronounce that the constructor should be explicit and the users should write `instance.send(Message(t_cur))`. Make sense?

LourensVeen:
Oh, I just spotted a subtle issue. The Fortran API is generated, and you edited the generated files rather than editing the generator and regenerating. I'll fix that.

PR #157, opened 2023-01-13
Documentation for checkpointing functionality

### LINKED ISSUE #139: Bullet lists are not correctly rendered
Unordered lists in the documentation are currently rendered without the bullets displayed. See for example: https://muscle3.readthedocs.io/en/latest/contributing.html#make-an-issue

This was an [issue in sphinx_rtd_theme](https://github.com/readthedocs/sphinx_rtd_theme/issues/1115) which should be resolved.

However.. we're using an old version of `sphinx_rtd_theme` (0.5.1) due to dependency conflicts with `sphinx_tabs`:
- `sphinx_tabs` requires `docutils~=0.18.0`
- The latest version of `sphinx_rtd_theme` requires `docutils<0.18` so cannot be used.

Docutils 0.18 support is on the roadmap for the 1.2 release of `sphinx_rtd_theme`: https://github.com/readthedocs/sphinx_rtd_theme/issues/1322

The easiest thing to do is wait for this new release. Though I cannot find an expected release date for 1.2, above issue is planned in the current sprint (planned to complete next week).

maarten-ic:
Additionally, fixes #139


LourensVeen:
Looks good, I've fixed a few typos and such, but nothing major. Maybe we should consider reorganising things a bit at some point, and moving the part about defining checkpoints in with the rest of the description of how to run simulations. But we'll get to that at some point.

I really don't want to have the deep dive in the user documentation. How it works is something that the user should never  ever have to think about. They use the API, they define checkpoints, they get snapshots, and that's it. So I've cut that off and moved it to a newly wrapped up section for MUSCLE3 developers. And maybe one day we'll write a paper about it, but the users need to worry about their science, not about checkpointing implementations.

I also had some issues with the Python API docs not rendering on RTD, but they seem to have magically disappeared. Oh well, time to merge this!

PR #183, opened 2023-01-31
Prototype to fix #182

### LINKED ISSUE #182: Program SEGFAULTS when linking against both libmuscle_mpi and libmuscle
When a C++ or Fortran program is targeting the `mpi` version of libmuscle, but links against both the mpi and regular shared library, SEGFAULTS or other undefined behaviour may happen.

When linking with `-lmuscle -lmuscle_mpi` (in that order!), the following happens:
- When constructing the instance, the call signature (with `MPI_Comm` and `root` arguments) is only provided by the `mpi` library, so the muscle instance is created by the `mpi` enabled version of the library
- When calling other methods of the public API, both `libmuscle.so` and `libmuscle_mpi.so` provide an implementation. Since `-lmuscle` was provided first when linking, the non-MPI implementation provided by `libmuscle.so` is used! Since the non-MPI library has a different memory layout of the Instance object (it doesn't have the MPI members), this will likely trigger a SEGFAULT, undefined behaviour, or more subtle bugs.

Suggested solution:
- Implement the MPI constructor signature in the non-MPI version (and vice-versa), and make them throw a runtime error.

maarten-ic:
@LourensVeen I couldn't mimick the MPI constructor in the non-MPI version of libmuscle without importing `mpi.h`, so I tried a different approach.

Please check if you agree, then I'll finish the PR by adding proper error messages.

### DIFF
--- a/libmuscle/cpp/src/libmuscle/instance.cpp
+++ b/libmuscle/cpp/src/libmuscle/instance.cpp
@@ -919,6 +919,11 @@ Instance::Instance(
                 , communicator, root
 #endif
                 ))
+#ifdef MUSCLE_ENABLE_MPI
+    , has_mpi_(true)
+#else
+    , has_mpi_(false)
+#endif
 {}
 
 Instance::~Instance() = default;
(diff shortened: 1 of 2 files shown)


maarten-ic:
See #182 for new proposed solution.

PR #184, opened 2023-02-01
Feature/flake8 update

### LINKED ISSUE #137: New flake8 gives errors
The newly released flake8 version 6 gives some spurious "X imported but unused" errors for some of our type annotations. This is because flake8 version 6 upgrades to pyflakes version 3, which removed support for them (see https://github.com/PyCQA/pyflakes/pull/684).

Type comments were introduced in PEP-484, and were the only way to write type annotations for variables until Python 3.6, when PEP-526 was implemented. So instead of `primes = []   # type: List[int]` we now get to write `primes: List[int] = []`.

We're kind of in the midst of things at the moment, so the short term solution is to specify `flake8<6` in tox.ini so the tests will run. Then we can go and remove all our type comments when the checkpointing/profiling/shared-mem dust settles down a little.

maarten-ic:
Fixes #137

### DIFF
--- a/docs/source/examples/python/interact_coupling.py
+++ b/docs/source/examples/python/interact_coupling.py
@@ -55,10 +55,10 @@ class DataCache:
 
         The cache starts out empty.
         """
-        self.t_cur = None       # type: Optional[float]
-        self.data_cur = None    # type: Optional[Any]
-        self.t_next = None      # type: Optional[float]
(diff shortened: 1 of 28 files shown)


LourensVeen (approved):
Nice, thank you! Now I just need to patch my brain to install the new syntax too...

PR #195, opened 2023-02-16
Feature/cpp instance flags

### LINKED ISSUE #181: should_save_final_snapshot has apply_overlay?
Didn't we decide to remove that option? There is now a flag on `Instance()` to use instead of passing this flag to `reuse_instance()`, and changing that is a small effort when adding support for the checkpointing API. So why is this still here?

maarten-ic:
Adds instance flags to C++ and Fortran APIs

- C++ `Instance()` constructors get two additional call signatures to allow backwards compatibility with v0.6.0.
  - `InstanceFlags` argument is added before MPI_COMM such that the non-MPI and MPI call signatures remain compatible.
  - Now there are 4 call signatures for the constructors, which is still manageable but a bit convoluted with the `#ifdef`s for MPI
- Fortran constructors are reworked to reduce repeated codes. The additional `InstanceFlags` **break backwards compatibility** for MPI models, but with a clear compiler error (at least on `gfortran`). See `changelog.rst`.
- C++/Fortran lose the `apply_overlay` argument to `reuse_instance()` (backwards incompatible, see `changelog.rst`).
- Python also loses the `apply_overlay` argument to `reuse_instance()` (was already deprecated in 0.6.0). Definitely closes #181.

### DIFF
--- a/docs/source/fortran_api.rst
+++ b/docs/source/fortran_api.rst
@@ -2150,23 +2150,23 @@ LIBMUSCLE_InstanceFlags
             You may not use any checkpointing API calls when this flag is not supplied.
 
     :f logical KEEPS_NO_STATE_FOR_NEXT_USE: Indicate this instance does not carry state
-            between iterations of the reuse loop.
+            between iterations of the reuse loop. Specifying this flag is equivalent to
+            :external:py:attr:`ymmsl.KeepsStateForNextUse.NO`.
 
-            This corresponds to :external:py:attr:`ymmsl.KeepsStateForNextUse.NO`.
-
-            If neither ``KEEPS_NO_STATE_FOR_NEXT_USE`` and
-            ``STATE_NOT_REQUIRED_FOR_NEXT_USE`` are supplied, this corresponds to
-            :external:py:attr:`ymmsl.KeepsStateForNextUse.NECESSARY`.
+            By default, (if neither ``KEEPS_NO_STATE_FOR_NEXT_USE`` nor
+            ``STATE_NOT_REQUIRED_FOR_NEXT_USE`` are provided), the instance is assumed
+            to keep state between reuses, and to require that state (equivalent to
+            :external:py:attr:`ymmsl.KeepsStateForNextUse.NECESSARY`).
 
     :f logical STATE_NOT_REQUIRED_FOR_NEXT_USE: Indicate this instance carries state
             between iterations of the reuse loop, however this state is not required
-            for restarting.
+            for restarting. Specifying this flag is equivalent to
+            :external:py:attr:`ymmsl.KeepsStateForNextUse.HELPFUL`.
 
-            This corresponds to :external:py:attr:`ymmsl.KeepsStateForNextUse.HELPFUL`.
-
-            If neither ``KEEPS_NO_STATE_FOR_NEXT_USE`` and
-            ``STATE_NOT_REQUIRED_FOR_NEXT_USE`` are supplied, this corresponds to
-            :external:py:attr:`ymmsl.KeepsStateForNextUse.NECESSARY`.
+            By default, (if neither ``KEEPS_NO_STATE_FOR_NEXT_USE`` nor
+            ``STATE_NOT_REQUIRED_FOR_NEXT_USE`` are provided), the instance is assumed
+            to keep state between reuses, and to require that state (equivalent to
+            :external:py:attr:`ymmsl.KeepsStateForNextUse.NECESSARY`).
 
 
 Namespace YMMSL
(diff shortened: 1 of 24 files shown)


docs/source/fortran_api.rst
```diff
@@ -2185,6 +2116,59 @@ LIBMUSCLE_Instance
...
+            supports checkpointing.
+
+            You may not use any checkpointing API calls when this flag is not supplied.
+
+    :f logical KEEPS_NO_STATE_FOR_NEXT_USE: Indicate this instance does not carry state
+            between iterations of the reuse loop.
+
+            This corresponds to :external:py:attr:`ymmsl.KeepsStateForNextUse.NO`.
+
+            If neither ``KEEPS_NO_STATE_FOR_NEXT_USE`` and
```
LourensVeen, line 2157: "neither" ... "nor"

Also, I didn't get what "this corresponds to" means on first reading. Maybe turn it around, "to get the equivalent of KeepsStateForNextUse.NECESSARY, specify neither KEEPS_NO_STATE_FOR_NEXT_USE nor STATE_NOT_REQUIRED_FOR_NEXT_USE"? Or "By default, the instance is assumed to keep state between reuses, and to require that state (equivalent to ymmsl.KeepsStateForNextUse.NECESSARY)"?

libmuscle/cpp/src/libmuscle/instance.cpp
```diff
@@ -668,13 +671,13 @@ bool Instance::Impl::receive_settings_() {
 /* Pre-receive on the given port and slot, if any.
  */
 void Instance::Impl::pre_receive_(
-        std::string const & port_name, Optional<int> slot,
-        bool apply_overlay) {
+        std::string const & port_name, Optional<int> slot) {
```
LourensVeen, line 674: This is nice without the explicit apply_overlay being passed everywhere!

libmuscle/cpp/src/libmuscle/instance.cpp
```diff
@@ -923,7 +926,40 @@ Instance::Instance(
...
+#ifdef MUSCLE_ENABLE_MPI
+                , communicator, root
+#endif
+                ))
+{}
+
+Instance::Instance(
+        int argc, char const * const argv[],
+        InstanceFlags flags
+#ifdef MUSCLE_ENABLE_MPI
```
LourensVeen, line 939: We could potentially make two macros like this and save some repetition, but I'm not sure it would actually make things more readable, what do you think?

```
#ifdef MUSCLE_ENABLE_MPI
#define MPI_ARGS_DECL , MPI_Comm const & communicator, int root
#define MPI_ARGS_CALL , communicator, root
#else
#define MPI_ARGS_DECL
#define MPI_ARGS_CALL
#endif
```
and then have
```
Instance::Instance(
        int argc, char const * const argv[],
        PortsDescription const & ports,
        InstanceFlags flags
        MPI_ARGS_DECL
        )
    : pimpl_(new Impl(
                argc, argv, ports, flags MPI_ARGS_CALL))
{}
```
```

File: libmuscle/fortran/src/libmuscle/libmuscle.f90 [hunk 4]
```diff
@@ -16505,11 +16484,13 @@ subroutine LIBMUSCLE_Message_unset_settings( &
             self%ptr)
     end subroutine LIBMUSCLE_Message_unset_settings
 
-    type(LIBMUSCLE_Instance) function LIBMUSCLE_Instance_create_autoports()
+    type(LIBMUSCLE_Instance) function LIBMUSCLE_Instance_create(ports, flags)
         implicit none
 
-        integer :: num_args, i, arg_len
-        integer (c_intptr_t) :: cla
+        type(LIBMUSCLE_PortsDescription), intent(in), optional :: ports
```
LourensVeen, line 16490: Ah, optional! I was thinking in C I guess, and never considered it. Nice!

libmuscle/fortran/src/libmuscle/tests/assert.f90
```diff
@@ -7,7 +7,7 @@ subroutine assert_true(x)
 
         if (.not. x) then
             print *, 'Assertion failed'
-            stop
+            stop 1
```
LourensVeen, line 10: Ha! Yeah, makes sense :smile:

LourensVeen (approved):
More features with less code! Nice! I'm happy to merge this as is, but I'll let you have a look at the comments and see if you want to play with macros some more first :smile:

libmuscle/cpp/src/libmuscle/instance.cpp
maarten-ic: I've applied it in `instance.cpp` and now all Instance constructor implementations fit on a single screen (including the macro definitions).

Although it makes the calls harder to read, I still think it's an improvement compared to all `#ifdef`s that were previously littering the place 😄

PR #209, opened 2023-03-27
Feature/oo fortran

maarten-ic:


scripts/make_libmuscle_api.py
```diff
@@ -68,43 +66,44 @@ def __init__(self, with_names: bool) -> None:
...
-                        '        names_v.push_back(std::string(index_name_6, index_name_6_size));\n'
-                        '    if (data_array_ndims >= 7u)\n'
-                        '        names_v.push_back(std::string(index_name_7, index_name_7_size));\n'
-                        '\n'
-                        '    Data * result = new Data(Data::grid(\n'
-                        '            data_array_p, data_array_shape_v,\n'
-                        '            names_v, libmuscle::StorageOrder::first_adjacent));\n'
-                        '    return reinterpret_cast<std::intptr_t>(result);\n'
-                        '}}\n\n').format(typ.tname(), typ.fc_cpp_type())
+                fc_override = dedent("""\
```
LourensVeen, line 69: I don't mind reformatting like this, and with the dedent I think it's better than what was there, but could we make that a separate PR next time? It's very difficult to see what is a simple reformat and what is a change in functionality when they're mixed together.

scripts/make_libmuscle_api.py
```diff
@@ -1026,13 +1041,14 @@ def __copy__(self) -> 'Elements':
             "KEEPS_NO_STATE_FOR_NEXT_USE",
             "STATE_NOT_REQUIRED_FOR_NEXT_USE"])
 
+
 instance_desc = Class(
-        'Instance', None, [instance_constructor] + [
+        'Instance', None, [cast(Member, instance_constructor)] + [
```
LourensVeen, line 1046: Why do we need a cast? It's a `Constructor`, which is a subclass of `Member`, right?

scripts/api_generator.py
```diff
@@ -2399,11 +2628,14 @@ def __init__(self, name: str, public: Optional[bool], prefix: str,
         self.prefix = prefix
         self.enums = enums
         self.classes = classes
+        self.flags = flags
```
LourensVeen, line 2631: Why are these separate now? They're still things in the namespace, right? And it seems like we do the same thing for them as for classes. If it's only for typing, I think it's better to combine them in a `members: List[Union[Class, Flags]]`. Actually, maybe enums should also be in that list, unless they need to be treated specially.

scripts/convert_fortran_source.py
```diff
@@ -0,0 +1,44 @@
+import re
```
LourensVeen, line 1: Ah! That explains :smile: I think `YMMSL_Settings` is missing from this script!

scripts/make_libmuscle_api.py
LourensVeen: For example, I noticed the change from push_back to emplace_back in the generated code (nice!), but it took me forever to find where the generator had changed, because I read right over it when reviewing this file, and couldn't find it in the generator code itself (because it wasn't there, obviously).

LourensVeen (commented):
Nice work, hopefully the last big one :smile:. I left a few comments.

scripts/make_libmuscle_api.py
maarten-ic: Yep, makes sense, sorry!

scripts/make_libmuscle_api.py
maarten-ic: Because `[instance_constructor]` is of type `List[Constructor]` and you're not allowed to add a `List[Member]` to it:

```
scripts/make_libmuscle_api.py:1046: error: Argument 3 to "Class" has incompatible type "List[Constructor]"; expected "List[Member]"
scripts/make_libmuscle_api.py:1046: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
scripts/make_libmuscle_api.py:1046: note: Consider using "Sequence" instead, which is covariant
scripts/make_libmuscle_api.py:1047: error: List comprehension has incompatible type List[Member]; expected List[Constructor]
```

Therefore I explicitly cast it to `Member` before putting it in the list.

scripts/api_generator.py
```diff
@@ -56,21 +101,80 @@ class is in.
...
 
-    def f_aux_variables(self) -> List[Tuple[str, str]]:
-        return list()
+        Used when generating a name for instantiated templates."""
 
-    def f_chain_arg(self) -> str:
-        return self.name
+    @abc.abstractmethod
+    def fc_cpp_type(self) -> str:
+        ...
```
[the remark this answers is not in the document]
maarten-ic: The Python doc for the `abc` module is using Ellipsis in the examples: https://docs.python.org/3/library/abc.html

You're not allowed to instantiate an abstract class (any base class that does not override all abstract methods), so it doesn't really matter.

I'm happy to update it to `raise NotImplementedError`, let me know if you want that.

scripts/api_generator.py
maarten-ic: Check, I've introduced a base class for flags, enums and classes.

scripts/convert_fortran_source.py
maarten-ic: Updated!

scripts/api_generator.py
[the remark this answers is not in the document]
LourensVeen: Okay, interesting. Maybe it's changed now that we have mypy and don't need to do this check at runtime anymore? At least that's the case here, so if the docs do that, then let's just follow suit.

scripts/make_libmuscle_api.py
LourensVeen: Ah of course! Makes sense now.

scripts/api_generator.py
LourensVeen: Oh, that's even nicer than a union, thanks!

LourensVeen (approved):
I like how this worked out, some nice deduplication. Let's merge this!

PR #212, opened 2023-03-29
Feature/run all check consistent

### LINKED ISSUE #145: Manager should catch and print loading errors
In `muscle_manager.py`, we load the yMMSL files passed on the command line. This may raise a RecognitionError if there's an error in a file. This currently causes the manager to crash. Instead, the exception should be caught, printed nicely to the screen, and then we should exit.

### LINKED ISSUE #161: Improve error message when component implementation cannot be found
## Reproduction scenario:

1. Update file `docs/source/examples/rd_python.ymmsl`, change macro/implementation to `diffusion_pythonX` (or something else for which no entry exists in the implementations)
2. Run muscle_manager (e.g. `make test_examples`)
3. Observe error message:

```
  File "/home/maarten/projects/muscle3/docs/source/examples/python/build/venv/bin/muscle_manager", line 8, in <module>
    sys.exit(manage_simulation())
  File "/home/maarten/projects/muscle3/docs/source/examples/python/build/venv/lib/python3.8/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/maarten/projects/muscle3/docs/source/examples/python/build/venv/lib/python3.8/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/maarten/projects/muscle3/docs/source/examples/python/build/venv/lib/python3.8/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/maarten/projects/muscle3/docs/source/examples/python/build/venv/lib/python3.8/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/maarten/projects/muscle3/docs/source/examples/python/build/venv/lib/python3.8/site-packages/muscle3/muscle_manager.py", line 74, in manage_simulation
    manager.start_instances()
  File "/home/maarten/projects/muscle3/libmuscle/python/libmuscle/manager/manager.py", line 98, in start_instances
    self._instance_manager.start_all()
  File "/home/maarten/projects/muscle3/libmuscle/python/libmuscle/manager/instance_manager.py", line 97, in start_all
    allocations = self._planner.allocate_all(self._configuration)
  File "/home/maarten/projects/muscle3/libmuscle/python/libmuscle/planner/planner.py", line 467, in allocate_all
    exclusive = {
  File "/home/maarten/projects/muscle3/libmuscle/python/libmuscle/planner/planner.py", line 470, in <setcomp>
    not implementations[c.implementation].can_share_resources)}
KeyError: Reference("diffusion_pythonX")
```

## Expected behaviour

A clear error message that the configuration is not correct (e.g. "Implementation 'diffusion_pythonX' of component 'macro' is not defined in the yMMSL configuration.")

### LINKED ISSUE #168: Check consistency of execution_model and resources
If you specify `openmpi` or `intelmpi` for the `execution_model` but also specify `threads` rather than `mpi_processes`, you should get an error, but you don't.

### LINKED ISSUE #206: Manager should print only error message if the yMMSL file(s) don't load correctly
Right now you get a whole backtrace with the error at the bottom, which is confusing and unnecessary. We should catch the exception and print a message saying that there was a problem reading only the error. Note that we can have exceptions from ruamel.yaml, YAtiML, and ymmsl-python.

maarten-ic:
Fixes #161, #168, #206 

Depends on: https://github.com/multiscale/ymmsl-python/pull/20

Related PR: https://github.com/multiscale/ymmsl-python/pull/21

### DIFF
--- a/libmuscle/python/libmuscle/manager/manager.py
+++ b/libmuscle/python/libmuscle/manager/manager.py
@@ -96,6 +96,7 @@ class Manager:
                     'For MUSCLE3 to be able to start instances, the'
                     ' configuration must contain a model, implementations,'
                     ' and resources. Please make sure they are all there.')
+        self._configuration.as_configuration().check_consistent()
         try:
             self._instance_manager.start_all()
         except:     # noqa
--- a/libmuscle/python/libmuscle/manager/manager.py
+++ b/libmuscle/python/libmuscle/manager/manager.py
@@ -101,7 +101,6 @@ class Manager:
             _logger.error(message)
             raise RuntimeError(message)
         try:
-            self._configuration.as_configuration().check_consistent()
             self._instance_manager.start_all()
         except:     # noqa
             _logger.error('An error occurred while starting the components:')
--- a/libmuscle/python/libmuscle/manager/manager.py
+++ b/libmuscle/python/libmuscle/manager/manager.py
@@ -96,8 +96,8 @@ class Manager:
                     'For MUSCLE3 to be able to start instances, the'
                     ' configuration must contain a model, implementations,'
                     ' and resources. Please make sure they are all there.')
-        self._configuration.as_configuration().check_consistent()
         try:
+            self._configuration.as_configuration().check_consistent()
             self._instance_manager.start_all()
         except:     # noqa
             _logger.error('An error occurred while starting the components:')
--- a/libmuscle/python/libmuscle/manager/manager.py
+++ b/libmuscle/python/libmuscle/manager/manager.py
@@ -108,7 +108,8 @@ class Manager:
 
     def stop(self) -> None:
         """Shuts down the manager."""
-        # self._server.stop()
+        if self._instance_manager:
+            self._instance_manager.shutdown()
         self._server.stop()
         self._snapshot_registry.shutdown()
         self._snapshot_registry.join()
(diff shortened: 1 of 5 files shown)


libmuscle/python/libmuscle/manager/manager.py
```diff
@@ -90,13 +90,18 @@ def get_server_location(self) -> str:
...
         if not self._instance_manager:
-            raise RuntimeError(
+            message = (
                     'For MUSCLE3 to be able to start instances, the'
                     ' configuration must contain a model, implementations,'
                     ' and resources. Please make sure they are all there.')
+            _logger.error(message)
+            raise RuntimeError(message)
         try:
+            self._configuration.as_configuration().check_consistent()
```
LourensVeen, line 104: Why would we do this here? Checking the configuration doesn't have anything to do with starting the instances. Sure, we need a valid configuration, but we're already doing some checks in `muscle_manager.py`, so why not call `check_consistent()` there as well?

LourensVeen (changes requested):
Looks good overall, one comment though that I would like to see fixed. Thanks!

LourensVeen (approved):
Great, nice also to have the check in case we don't use `start_all`, especially considering that we actually have a user doing that at the moment.

PR #215, opened 2023-03-29
Update integration test

maarten-ic:
This test works by taking snapshots based on wallclock_time. This is less reliable and failed every now and then in CI.

Updated the test, hopefully it doesn't fail anymore in CI after this.

### DIFF
--- a/integration_test/test_snapshot_complex_coupling.py
+++ b/integration_test/test_snapshot_complex_coupling.py
@@ -149,7 +149,7 @@ model:
 
 settings:
   dt: 1.234
-  t_max: 2.0  # seconds
+  t_max: 1.8  # seconds
   cacheA.cache_valid: [2.0, 5.0]
   cacheB.cache_valid: [3.0, 8.0]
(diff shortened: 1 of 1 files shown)


LourensVeen:
I don't get it. Is this just about statistics, a shorter run leaves fewer opportunities for something to go wrong?

maarten-ic:
Previously the components were supposed to run for 2.0 seconds, and take a snapshot every 0.5 second AND take a snapshot at the end. This was supposed to lead to 5 snapshots (at 0.5, 1.0, 1.5 and 2.0 seconds and the final one), which it always did on my laptop.

However, on CI, it sometimes made only 4 snapshots. My hunch is that the 2.0 second one was not always taken (or being the same as the final one) in CI, so now I've let it run for only 1.8 seconds so the 2.0 second snapshot should never be taken. Hopefully that fixes the random CI failures 🤞

LourensVeen (approved):
Okay, well, at least it won't make it worse!

Issue #102, opened 2022-08-15
Editable pip install is not working correctly


maarten-ic:
When following [these instructions on the documentation](https://muscle3.readthedocs.io/en/latest/contributing.html#install-the-tools), the package is installed through `pip install -e .[dev]`.

Unfortunately, editable installs are not supported with a custom `pkg_dir` in setup.py. See for instance this ticket: https://github.com/pypa/pip/issues/3160
Eventually you'll get ModuleNotFoundErrors like below:

```bash
$ muscle_manager --help
Traceback (most recent call last):
  File "/home/[...]/venv_muscle3/bin/muscle_manager", line 33, in <module>
    sys.exit(load_entry_point('muscle3', 'console_scripts', 'muscle_manager')())
  File "/home/[...]/venv_muscle3/bin/muscle_manager", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/usr/lib/python3.8/importlib/metadata.py", line 77, in load
    module = import_module(match.group('module'))
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 848, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/[...]/projects/muscle3/muscle3/muscle_manager.py", line 10, in <module>
    from libmuscle.manager.manager import Manager
ModuleNotFoundError: No module named 'libmuscle.manager'
```

Suggested immediate fix: update documentation to not use `pip install -e`.

If I have better suggestions I'll update this issue.

DaanVanVugt:
Relevant discussion at https://github.com/pypa/setuptools/issues/230

DaanVanVugt:
Adding the `--use-pep517` flag to get `pip install --use-pep517 -e .[dev]` resolves the problem. I believe PEP517 is enabled by default when using pyproject.toml, so at that point the flag can be removed again.

maarten-ic:
With #348, editable installs are working out-of-the-box

PR #218, opened 2023-04-07
update docs on editable mode install

DaanVanVugt:
Improves docs to prevent people from running into #102


LourensVeen (approved):
Nice, I guess this fix may not actually have been available when the issue was created :smile:.

Issue #163, opened 2023-01-24
Manager doesn't detect instance that quits immediately


LourensVeen:
If an instance is started, and then finishes immediately with a 0 exit code and without ever registering, then the manager does not detect that anything is wrong, and will just sit there waiting for the instance to appear.

PR #223, opened 2023-04-16
Handle instances that quit with 0 before registering

LourensVeen:
This addresses #163 by making the InstanceManager check with the InstanceRegistry that an instance which quit with exit code 0 did register before doing so. If it didn't, that's an error and handled accordingly by shutting down the rest of the simulation and logging an error.

This needs #222 to be merged first.

### DIFF
--- a/libmuscle/python/libmuscle/manager/instance_manager.py
+++ b/libmuscle/python/libmuscle/manager/instance_manager.py
@@ -7,6 +7,7 @@ import queue
 
 from ymmsl import Configuration
 
+from libmuscle.manager.instance_registry import InstanceRegistry
 from libmuscle.manager.instantiator import (
         CancelAllRequest, CrashedResult, InstantiatorRequest,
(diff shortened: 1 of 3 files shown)


maarten-ic (approved):
Great improvement!

FYI: you can automatically link issues by using one of these keywords: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

"Addresses" is not one of them, if you had used "Fixes" or any of the other options it would automatically be linked. I've now linked it manually 😄

LourensVeen:
I'm using "addresses" intentionally :smile:. I don't like closing an issue reported by a user before the fix for it is actually in the user's hands. Just merging it into `develop` doesn't actually fix their issue after all.

So I have a "Ready for release" label that I apply after it's merged, and the issue isn't closed until a new version has been released with the fix in it, preferably by the original submitter of the issue after confirming that it works for them now.

PR #227, opened 2023-04-24
Update test_snapshot_complex_coupling

maarten-ic:
Remove all checks for a specific amount of snapshots: Snapshotting based on wallclock time is too dependent on circumstances to reliably assert a specific amount of snapshots.


LourensVeen (approved):
That should save some CI failures, good.

Issue #78, opened 2020-10-28
Profiling functionality unfinished


LourensVeen:
In a MUSCLE3 simulation, communication between instances is done via libmuscle, as is waiting for other instances to produce the required data. Given this, we can implement _semi-intrusive profiling_, which records performance information on communication and allows the user to determine which model takes up most of the time, and how much time instances spent on communicating, waiting for data, and everything else (i.e. computing). This can be used by the user to guide performance improvements, but also for a future semi-intrusive optimiser.

There is some data collection code on the Python side, but it's unfinished.

- [x] Extend Python-side data collection
- [x] Add C++/Fortran support
- [x] Save events to SQLite database
- [x] Record message numbers, which were added with checkpointing and help with analysing the data
- [x] Allow disabling via a `muscle_profile_level` setting, `none` or `all` (1)
- [x] Batch the sending of the events based on time rather than count
- [x] Use a background thread to save the data
- [x] Add basic analysis functionality
- [x] Do some benchmarking and compare with e.g. perf or gprof output to make sure the results make sense
- [x] Document

(1) This will allow disabling profiling for large simulations, where the manager could get overwhelmed by the volume of events being generated. Eventually, we may want to implement a `brief` setting that summarises the data a bit within each instance, to limit traffic but still give useful data. Some user feedback on what information they need would help with that.

LourensVeen:
Released in 0.7.0.

PR #232, opened 2023-06-17
Finish initial profiling functionality

LourensVeen:
This completes the initial profiling functionality described in #78.

Maarten and Daan: no need to review as it's outside of our project, but you were interested in playing with it and if you have any comments feel free to add. I decided to keep the timeline plot even if it isn't perfect, it's just too nice not to include :smile:.


libmuscle/python/libmuscle/manager/profile_database.py
```diff
@@ -10,27 +16,419 @@ class ProfileDatabase:
...
+        total_times = [(stop_run[i] - start_run[i]) * 1e-9 for i in instances]
+        comm_times = [
+                (
+                    (comm[i] if i in comm else 0) -
+                    (wait[i] if i in wait else 0)
+                ) * 1e-9
+                for i in instances]
+        wait_times = [(wait[i] if i in wait else 0) * 1e-9 for i in instances]
+        run_times = [
+                t - c - w
```
maarten-ic, line 140: This subtracts `wait[i]` twice `if i in wait`, is that correct?

libmuscle/python/libmuscle/manager/profile_database.py
```diff
@@ -10,27 +16,419 @@ class ProfileDatabase:
...
+                for t, c, w in zip(total_times, comm_times, wait_times)]
+
+        return instances, run_times, comm_times, wait_times
+
+    def resource_stats(self) -> Dict[str, Dict[str, float]]:
+        """Calculate per-core statistics.
```
maarten-ic, line 146: This assumes that a multi-core instance is distributing the workload evenly on all cores, right?

libmuscle/python/libmuscle/manager/profile_store.py
```diff
@@ -74,16 +138,51 @@ def to_tuple(e: ProfileEvent) -> Record:
...
-                map(to_tuple, events))
-        cur.execute("COMMIT")
         cur.close()
+        super().close()
+
+    def _get_instance_oid(self, cur: Cursor, instance_id: Reference) -> int:
+        """Get the oid for a given instance.
+
+        Inserts instance into the database and returns the new oid if
+        needed. Does not start its own transaction.
```
maarten-ic, line 173: I don't see an INSERT statement in this method?

libmuscle/python/libmuscle/manager/test/test_profile_database.py
```diff
@@ -0,0 +1,137 @@
...
+from ymmsl import Operator, Port, Reference
+
+import pytest
+
+from pathlib import Path
+
+
+@pytest.fixture
+def db_file(tmp_path) -> Path:
+    with patch('libmuscle.manager.profile_store._SYNCHED', True):
```
maarten-ic, line 18: For context (no need to change anything): you can also use [pytest's `monkeypatch.setattr`](https://docs.pytest.org/en/6.2.x/monkeypatch.html). Then you don't need to put all the logic inside a `with` statement.

docs/source/profiling.rst
```diff
@@ -0,0 +1,498 @@
...
+Profiling is enabled by default, so you do not have to do anything to get
+MUSCLE3 to produce profiling data. For typical coupled simulations consisting of
+only a few instances that do not communicate very often, this should not affect
+the simulation performance. If you have a simulation with very many instances
+which communicate intensively, then the overhead of sending all that information
+to the central database can become significant.
+
+If that is the case, then you should do your profiling with a smaller number of
+instances, and disable profiling for the large run to avoid the overhead.
+Profiling can be disabled by setting the setting ``muscle_profile_level`` to
```
maarten-ic, line 42: Might be good to include an overview of `special` settings somewhere in the documentation. Not for this PR (you may create an issue if you agree). The other special settings I can think of is the remote log level, but perhaps there are more already.

maarten-ic (approved):
Very nice work!

DaanVanVugt (approved):
Looks amazing!

docs/source/examples/benchmark_implementations.ymmsl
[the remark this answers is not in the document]
LourensVeen: Ah, it was, but I guess I messed up a git command somewhere. You're completely right.

libmuscle/python/libmuscle/manager/profile_database.py
LourensVeen: Yes, essentially it takes those resources as occupied during that time, and whatever the model code is doing internally is its concern, and for some other software to measure. Although I'm sure people will eventually want some documentation on how to profile submodels within a MUSCLE3 simulation...

libmuscle/python/libmuscle/manager/profile_store.py
LourensVeen: Ah, this changed and the comment is obsolete. Will fix.

docs/source/profiling.rst
```diff
@@ -0,0 +1,498 @@
...
+If you are running on HPC, then you most likely have a fixed budget of
+core hours or system billing units (SBUs), and reducing SBU usage may be of
+interest. In the above scenario, both models are actually candidates for
+optimisation in this case. If you can speed up the parallel model so that it can
+run as fast on fewer cores then that potentially saves many SBUs. But speeding
+up the single-thread model also helps, because even if the cores that the
+parallel model runs on aren't used, they're still allocated and you're still
+paying for them.
+
+If all your models spend most of their time waiting, then your models do
```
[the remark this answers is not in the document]
LourensVeen: Hmm, good point. The core issue here is that the total time spent computing is low compared to the time spent sending and receiving, because that means there's a lot of communication overhead. I'll rephrase.

libmuscle/python/libmuscle/manager/profile_database.py
```diff
@@ -10,27 +16,419 @@ class ProfileDatabase:
...
+            return cast(Tuple[int, int, int], result)
+
+        cur = self._get_cursor()
+        sum1_high, sum1_low, count1 = get_sum_count(
+                cur, etype, timestamp, instance, port, slot)
+        sum2_high, sum2_low, count2 = get_sum_count(
+                cur, etype2, timestamp2, instance, port2, slot2)
+        cur.close()
+
+        if count1 != count2:
```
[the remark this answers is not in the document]
LourensVeen: No, we could have one series of events only taking place in the beginning of the simulation, and the other one throughout the whole thing. Calculating an average would give very incorrect results.

The solution in that case may be to merge on message number, but those may also not match. I think we need some practical cases to see what makes sense to support.

libmuscle/python/libmuscle/manager/profile_database.py
```diff
@@ -10,27 +16,419 @@ class ProfileDatabase:
...
+            with ProfileDatabase('performance.sqlite') as db:
+                # use db here
         """
         # If the thread never served a request, then we don't have a
         # connection.
         if hasattr(self._local, 'conn'):
             self._local.conn.close()
 
+    def instance_stats(
+            self) -> Tuple[List[str], List[float], List[float], List[float]]:
```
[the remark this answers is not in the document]
LourensVeen: Ah, interesting. I knew about collections.named_tuple but not this one. It's nice, but it's also an extra class that we have to explain to people. Especially for this kind of data analysis stuff I think plain types is probably better. Or a Pandas dataframe, but I don't want to depend on that...

libmuscle/python/libmuscle/manager/profile_database.py
LourensVeen: RECEIVE events include the waiting period, so they're broken down into RECEIVE_WAIT, RECEIVE_TRANSFER and RECEIVE_DECODE. So this takes the total time, subtracts the communication and waiting time, adds the waiting time back on, then subtracts it again. Convoluted, but correct :smile:

libmuscle/python/libmuscle/manager/test/test_profile_database.py
LourensVeen: Oh, interesting. I'm going to have a look at that when I have more time.

docs/source/profiling.rst
LourensVeen: I think it's only these two so far, but good idea. #235

Issue #216, opened 2023-03-31
Manager doesn't flush print of its ip:port string


peter-t-fox:
If starting the manager and models manually from a script, the script needs to capture the `tcp:<ip>:<port>` string printed by the manager, and pass it to the models in their environment.

However, the manager's `stdout` is not flushed, and so this is difficult e.g.

```
$ muscle_manager --no-start-all configuration.ymmsl > stdout.txt &
[1] <pid>
$ cat stdout.txt
$ 
```

We can force a flush by using the PYTHONUNBUFFERED environment variable:

```
$ PYTHONUNBUFFERED=1 muscle_manager --no-start-all configuration.ymmsl > stdout.txt &
[2] <pid>
$ cat stdout.txt 
tcp:127.0.0.1:9000
```

or by using `python -u` to start the manager.

Since the network information printed to `stdout` is needed for use in the case where the user is starting each of the components manually, it might be helpful to support that use case by updating the [line](https://github.com/multiscale/muscle3/blob/dfd7d64aa7be7c6d8fea2d1d18e5c2f1e42662c8/muscle3/muscle_manager.py#L80):

```
print(manager.get_server_location())
```

to

```
print(manager.get_server_location(), flush=True)
```

This way the user doesn't have to manually force a flush of the buffer.

LourensVeen:
Hi Peter, sorry for the slow reaction, and thanks for reporting!

I'll probably add a command line option to write the string to a file to avoid potential confusion with other output, but I'll keep this and make sure to flush the buffer as well.

peter-t-fox:
Great, thank you.

LourensVeen:
This was released with version 0.7.0. Could you please test, and if this fixes it for you, close this issue?

peter-t-fox:
Thanks! Sorry I didn't see this sooner: I don't seem to have notifications enabled. I am still working with version 0.6.0, but I should be able to test this soon. Thanks for making this change: it looks like it should do what I need.

peter-t-fox:
@LourensVeen I have switched to using MUSCLE3 0.7.0, and I have had an opportunity to test this. It's now working as I would expect: the addition of `flush=True` means I don't have to to force the manager to print the IP address via `PYTHONUNBUFFERED` or similar.

So I think this issue can now be closed: thanks for you help!

PR #238, opened 2023-06-20
Add --location-file option to muscle_manager (#216)

LourensVeen:
This adds a new option to the `muscle_manager` program to specify that the manager's network location should be written to a given file. If the option is given, the location will not be printed to standard output.

If the option is not given, the location will be printed to standard output as before, except that we now flush the stream so as to make sure it actually ends up with the reader in a finite amount of time.

Addresses #216


LourensVeen:
Failed tests are due to a compatibility issue with mypy and typing_extensions and are not related to this change, which can be merged safely.

muscle3/muscle_manager.py
```diff
@@ -33,6 +33,25 @@
...
+@click.option(
+        '--location-file', nargs=1, type=click.Path(
+            exists=False, file_okay=True, dir_okay=True, readable=False,
+            allow_dash=True),
+        help=(
+            'File to write the network location of the manager to if'
+            ' --start-all is not specified. If a relative path is given,'
+            ' then it will be resolved relative to the directory in which'
+            ' the manager was started.'
+            ''
```
maarten-ic, line 45: These newlines are not preserved in the output of `muscle_manager --help` , not sure if that is your intention?

muscle3/muscle_manager.py
```diff
@@ -105,7 +125,13 @@ def manage_simulation(
             manager = Manager(configuration, None, log_level)
         else:
             manager = Manager(configuration, run_dir_obj, log_level)
-        print(manager.get_server_location())
+
+        server_location = manager.get_server_location()
+        if location_file is None:
+            print(server_location, flush=True)
+        else:
+            with Path(location_file).open('w') as f:
```
maarten-ic, line 133: You could do `Path(...).write_text(server_location)`: https://docs.python.org/3/library/pathlib.html#pathlib.Path.write_text

maarten-ic (approved):

muscle3/muscle_manager.py
LourensVeen: Sort of, I was mainly too lazy to look up how to do it. Turns out there is a way though, and you're right, so I fixed it.

muscle3/muscle_manager.py
LourensVeen: Ooooh, nice, didn't know about that one. Or rather, I guess I forgot about it, because I implemented it a few years ago :smile:: https://github.com/MD-Studio/cerulean/blob/e081f0242494890e26b5ec5e150187c1f130477f/cerulean/path.py#L447

Issue #247, opened 2023-06-28
After increasing the number of copies of my physics code from 10 to 20, I have started on seeing errors ...


DavidPCoster:
After increasing the number of copies of my physics code from 10 to 20, I have started on seeing errors ...

Traceback (most recent call last):
  File "/afs/eufus.eu/user/g/g2dpc/GIT/ets_paf/UQ/load_balancer.py", line 60, in <module>
    main()
  File "/afs/eufus.eu/user/g/g2dpc/GIT/ets_paf/UQ/load_balancer.py", line 34, in main
    while instance.reuse_instance():
  File "/pfs/scratch/g2dpc/GIT/ets_paf/UQ/fusion_env/lib/python3.10/site-packages/libmuscle/instance.py", line 232, in reuse_instance
    do_reuse = self._decide_reuse_instance()
  File "/pfs/scratch/g2dpc/GIT/ets_paf/UQ/fusion_env/lib/python3.10/site-packages/libmuscle/instance.py", line 851, in _decide_reuse_instance
    got_f_init_messages = self._pre_receive()
  File "/pfs/scratch/g2dpc/GIT/ets_paf/UQ/fusion_env/lib/python3.10/site-packages/libmuscle/instance.py", line 1018, in _pre_receive
    self.__pre_receive_f_init()
  File "/pfs/scratch/g2dpc/GIT/ets_paf/UQ/fusion_env/lib/python3.10/site-packages/libmuscle/instance.py", line 1084, in __pre_receive_f_init
    pre_receive(port_name, slot)
  File "/pfs/scratch/g2dpc/GIT/ets_paf/UQ/fusion_env/lib/python3.10/site-packages/libmuscle/instance.py", line 1061, in pre_receive
    msg, saved_until = self._communicator.receive_message(
  File "/pfs/scratch/g2dpc/GIT/ets_paf/UQ/fusion_env/lib/python3.10/site-packages/libmuscle/communicator.py", line 386, in receive_message
    _logger.debug('Received message on {}'.format(port_and_slot))
  File "/gw/switm/python/3.10/lib/python3.10/logging/__init__.py", line 1465, in debug
    self._log(DEBUG, msg, args, **kwargs)
  File "/gw/switm/python/3.10/lib/python3.10/logging/__init__.py", line 1624, in _log
    self.handle(record)
  File "/gw/switm/python/3.10/lib/python3.10/logging/__init__.py", line 1634, in handle
    self.callHandlers(record)
[...]

PR #256, opened 2023-07-06
Protect manager connection from race conditions in MMPClient

LourensVeen:
This adds locking around the connection to the manager, which ensures there are no race conditions between log messages being sent to the manager by the main thread, and profiling information being sent to the manager by the profiling communication background thread.


LourensVeen:
Addresses #247.

PR #299, opened 2024-08-13
Feature/deadlock detection

maarten-ic:
Add a deadlock detection mechanism to MUSCLE3.

New functionality:
- Instance sends a message to the manager when it is waiting for longer than `muscle_deadlock_receive_timeout` seconds (configurable in muscle settings)
- The manager reports deadlocks (a cycle of instances that are all waiting on each other) and shuts down the simulation

## TODO
- [x] Implement Instance logic in C++ as well
- [x] Add integration tests
- [x] Add documentation


libmuscle/python/libmuscle/manager/deadlock_detector.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def _process_queue_item(self, item: _QueueItem) -> None:
+        """Actually process a WAITING_FOR_RECEIVE[_DONE] request.
+
+        This method should be called inside the worker thread.
+        """
+        _logger.debug("Processing queue item: %s", item)
+        is_waiting, instance_id, peer_instance_id, port_name, slot = item
+        if is_waiting:
+            # Sanity checks, triggering this is a bug in the instance or the manager
```
LourensVeen, line 129: I would use an `assert` here, this is just what they're for.

libmuscle/python/libmuscle/manager/deadlock_detector.py
```diff
@@ -0,0 +1,220 @@
...
+                    "Instance %s was already waiting on a receive call. "
+                    "Did we miss a WAITING DONE event?",
+                    instance_id)
+            # Register that the instance is waiting
+            self._waiting_instances[instance_id] = peer_instance_id
+            self._waiting_instance_ports[instance_id] = (port_name, slot)
+            self._check_for_deadlock(instance_id)
+
+        else:
+            # Sanity checks, triggering these is a bug in the instance or the manager
```
LourensVeen, line 141: Same here.

libmuscle/python/libmuscle/manager/deadlock_detector.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def _check_for_deadlock(self, instance_id: str) -> None:
+        """Check if there is a cycle of waiting instances that involves this instance.
+        """
+        deadlock_instances = [instance_id]
+        cur_instance = instance_id
+        while cur_instance in self._waiting_instances:
+            cur_instance = self._waiting_instances[cur_instance]
+            if cur_instance == instance_id:
+                self._handle_potential_deadlock(deadlock_instances)
```
LourensVeen, line 178: I really like the progressive delegation from `_check_for_deadlock()` to `_handle_potential_deadlock()` to `_format_deadlock()` here. Nice!

libmuscle/python/libmuscle/manager/deadlock_detector.py
```diff
@@ -0,0 +1,220 @@
...
+            self._shutdown_time = time.monotonic() + self._wait_before_shutdown
+
+    def _format_deadlock(self, deadlock_instances: List[str]) -> str:
+        """Create and return formatted deadlock debug info.
+
+        Args:
+            deadlock_instances: list of instances waiting on eachother
+        """
+        num_instances = str(len(deadlock_instances))
+        lines = [f"The following {num_instances} instances are dead-locked:"]
```
LourensVeen, line 210: We have deadlocks everywhere, but here things are dead-locked. I'm not seeing any hyphens in the dictionary, and "dead-" can also be used an an adjective (e.g. "I was dead sure") so I'd go with "deadlocked" instead.

(A nit, a nit, I found a nit! *Dances happy nitpicking dance*)

libmuscle/python/libmuscle/manager/deadlock_detector.py
```diff
@@ -0,0 +1,220 @@
...
+        for i, instance in enumerate(deadlock_instances):
+            num = str(i+1).rjust(len(num_instances))
```
LourensVeen, line 212: Doesn't `str.format()` have a way to format fields that would make sense to use here?

libmuscle/python/libmuscle/manager/deadlock_detector.py
```diff
@@ -0,0 +1,220 @@
...
+
+            # Grab a new item from the queue, this raises Empty when timeout expires:
+            try:
+                item = self._queue.get(timeout=timeout)
+                if item is None:  # On shutdown, None is pushed to the queue
+                    return  # exit thread
+                self._process_queue_item(item)
+
+            except Empty:
+                # timeout expired and queue is empty, call shutdown callback
```
LourensVeen, line 75: Hold on wait a minute, what if everyone is happily computing, and that's why we're not getting any messages about waits? Or they're communicating but it's all going back and forth so fast that there's never anyone waiting long enough to send a message?

I think we need to check here if there are any deadlocks, and if not, continue the while loop.

libmuscle/python/libmuscle/manager/manager.py
```diff
@@ -121,6 +125,9 @@ def stop(self) -> None:
         """Shuts down the manager."""
         if self._instance_manager:
             self._instance_manager.shutdown()
+        self._deadlock_detector.shutdown()
+        # Note: don't join() deadlock detector, as this method may be called from the
+        # DeadlockDetector thread. join() would (ironically) deadlock the shutdown :)
```
LourensVeen, line 130: Now if only this were Java, then we could have a `class DeadlockDetectorShutdownDeadlockDetector` and it would be some kind of design pattern that we could put onto our CVs!

LourensVeen (changes requested):
Good start! I do have an architectural issue and an algorithmic one, and some small comments.

libmuscle/python/libmuscle/mcp/tcp_transport_client.py
```diff
@@ -66,13 +76,41 @@ def call(self, request: bytes) -> Tuple[bytes, ProfileData]:
...
+        Args:
+            timeout: timeout in seconds
+
+        Returns:
+            True if the socket is ready for receiving data, False otherwise.
+        """
+        if self._poll_obj is not None:
+            ready = self._poll_obj.poll(timeout * 1000)  # poll timeout is in ms
+        else:
+            # Fallback to select()
```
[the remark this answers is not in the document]
maarten-ic: Yes, `select()` on Linux has the following limitation (from `man select`):

```
WARNING:  select() can monitor only file descriptors numbers that are less than FD_SETSIZE (1024)—an unreason‐
ably low limit for many modern applications—and this limitation will  not  change.   All  modern  applications
should instead use poll(2) or epoll(7), which do not suffer this limitation.
```

The select() fallback is there to allow Python-on-Windows muscle3 (where `poll()` is not available). I'm fine to remove the fallback if you don't want to support Windows :)

libmuscle/python/libmuscle/manager/deadlock_detector.py
maarten-ic: If everyone is happily communicating then they should have sent a `WAITING_FOR_RECEIVE_DONE` message.
This message would be processed before the timeout expires, and if there are no deadlocks, there is no new timeout set.

It's not the most intuitive logic though, so I'll add a comment and put an assert double-checking for deadlocks here :)

libmuscle/python/libmuscle/manager/manager.py
maarten-ic: 😅 

I've actually found later in the Python docs that calling `Thread.join()` on your own thread raises a RuntimeError instead of deadlocking, so I doubly prefer that this is Python instead of Java 😜

libmuscle/python/libmuscle/mcp/tcp_transport_client.py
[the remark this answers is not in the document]
LourensVeen: Ah, I looked into this in quite a bit of detail when I wrote the TCP server, but I don't recall this one. And while Windows isn't a priority, it's nice if the Python tutorial using the runner works on Windows at least, for teaching. Now that psutil has replaced netifaces we're a bit closer to that (maybe it works actually, haven't tested), so it's good to have this. Thanks!

libmuscle/python/libmuscle/manager/deadlock_detector.py
LourensVeen: Ah, you're right, I didn't read carefully enough how the timeout is set. It makes sense, it'll only specify a timeout if we've already detected a deadlock. So PEBKAC here.

Maybe it would help if we introduced some terminology to more clearly distinguish between the different timeouts. There's a client-side timeout after which the manager is informed, and there's a server-side timeout to compensate for any potential delayed messages. Something with a grace period for the latter?

libmuscle/python/libmuscle/manager/manager.py
LourensVeen: Well, there go our careers...

libmuscle/python/libmuscle/manager/deadlock_detector.py
maarten-ic: It does, but then you'd have to do two string interpolations:
1. Create the format string to set a variable width of the field
2. Use that format string to format `i+1` with the correct width

This would look like the following:
```python
num = f"{{0:{len(num_instances)}d}}".format(i+1)
```

I think the current logic is more clear for our case. For a fixed width of the field (for example, always use 4 characters), `str.format()` would be my preferred approach as well 🙂

LourensVeen: Yuck! Okay, yes, good point!

libmuscle/python/libmuscle/communicator.py
```diff
@@ -60,6 +62,48 @@ def __init__(self, timestamp: float, next_timestamp: Optional[float] = None,
         self.settings = settings
 
 
+class RecvTimeoutHandler(TimeoutHandler):
```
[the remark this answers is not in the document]
maarten-ic: Yes makes sense. I've had to include a factory class as well to make this work, but it simplifies `communicator.py` so that's worth it IMO!

Note: we can't set it in `Communicator.__init__` because the timeout needs to have the settings received. To have the settings, we must have registered with the muscle manager. And `Instance._register` needs to have an initialized `Communicator`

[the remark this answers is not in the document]
LourensVeen: Ah, good point about __init__, I meant during initialisation in general, it's similar to setting the log levels.

libmuscle/cpp/src/libmuscle/receive_timeout_handler.cpp
```diff
@@ -9,16 +11,22 @@ ReceiveTimeoutHandler::ReceiveTimeoutHandler(
...
     , slot_(slot)
-    , timeout_(timeout) {}
+    , timeout_(timeout)
+    , num_timeout_(0) {}
 
 double ReceiveTimeoutHandler::get_timeout()
 {
-    return timeout_;
+    // Increase timeout by a factor 1.5 with every timeout we hit:
+    return timeout_ * std::pow(1.5, (double)num_timeout_);
```
LourensVeen, line 20: Should `num_timeout_` be incremented here? Or somewhere else?

libmuscle/python/libmuscle/manager/deadlock_detector.py
```diff
@@ -13,11 +13,11 @@ class DeadlockDetector(Thread):
     """The DeadlockDetector attempts to detect when multiple instances are stuck waiting
     for each other.
 
-    This class is responsible for handling WAITING_FOR_RECEIVE and
+    This class is responsible for handling WAITING_FOR_RECEIVE, IS_DEADLOCKED and
     WAITING_FOR_RECEIVE_DONE MMP messages, which are submitted by the MMPServer.
 
     When a deadlock is detected, the cycle of instances that is waiting on each other is
-    logged with FATAL severity. If this deadlock does not get resoled in
+    logged with FATAL severity. If this deadlock does not get resolved in
```
LourensVeen: Actually, it's only logged if it isn't resolved, right? That's how it should be, so the comment should be clarified I think.

LourensVeen (commented):
Okay, nice work again! I've left some comments, but it's looking good overall.

libmuscle/cpp/src/libmuscle/receive_timeout_handler.hpp
```diff
@@ -0,0 +1,40 @@
...
+ * another message when the message does arrive).
+ * 
+ * This is used by the manager to detect if the simulation is in a deadlock, where a
+ * cycle of instances is waiting on each other.
+ */
+class ReceiveTimeoutHandler : public mcp::TimeoutHandler {
+    public:
+        ReceiveTimeoutHandler(
+                MMPClient & manager,
+                std::string const & peer_instance,
```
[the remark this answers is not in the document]
maarten-ic: Sure thing, I've also updated it for the Python library to keep that consistent

libmuscle/cpp/src/libmuscle/receive_timeout_handler.cpp
maarten-ic: It's also in the C++ one, but only in added in a later commit when I found that the tests failed ;) https://github.com/multiscale/muscle3/pull/299/commits/f3f8e317acdb169a57ba398369bc6ef9ad52ec50#diff-7c147a535e0c6698203173d19156e7070c502df13dc79087da799e7bdc2113b6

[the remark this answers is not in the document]
maarten-ic: `(double)` can be omitted 👍

libmuscle/python/libmuscle/manager/deadlock_detector.py
maarten-ic: This is currently logged as soon as the manager identifies a cycle of waiting instances. I think this will never be resolved, but I'm not sure.

I can also update the logic to only print the deadlock cycle when any of the deadlocked instances calls `is_deadlocked` and starts shutting down. At that point we're sure that there was a deadlock and that the simulation will shut down.

libmuscle/cpp/src/libmuscle/receive_timeout_handler.cpp
LourensVeen: Ah, I reviewed one commit at a time. Good test of the tests and also of the review :smile: :.

LourensVeen (approved):

PR #300, opened 2024-08-20
Feature/unit test cleanup threads

maarten-ic:
This PR adds a fixture to check that no threads are left running after every test.

It also fixes every case where threads were left running, and thereby prevents the following `DeprecationWarning`:

```
DeprecationWarning: This process (pid=......) is multi-threaded, use of fork() may lead to deadlocks in the child.
```

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,10 @@
+import threading
+
+import pytest
+
+
+@pytest.fixture(autouse=True)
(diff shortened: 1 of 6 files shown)


LourensVeen (approved):
Wow, this was quite a mess, especially the queues as that's not just a test issue. Thanks for the fixes!

PR #306, opened 2024-11-12
Fix warning for invalid `muscle_local_log_level` in Python instance

maarten-ic:
Previously, an invalid `muscle_local_log_level` would be silently ignored in Python instances.

Restructured the error logic to match the one for `muscle_remote_log_level` and fixed the warning message itself.


LourensVeen (approved):
Doh! Too many KeyErrors...

Issue #315, opened 2025-05-16
Installing with long pathnames (headerpad)


sbr-shakibi:
Installing MUSCLE3 v0.8.0 on my MacOS device gives rise to errors similar to this:

```
error: /Library/Developer/CommandLineTools/usr/bin/install_name_tool: changing install names or rpaths can't be redone for: /Users/User_Name/Tissue-Simulation-Toolkit/lib/muscle3/muscle3/lib/libymmsl_d.dylib (for architecture arm64) because larger updated load commands do not fit (the program must be relinked, and you may need to use -headerpad or -headerpad_max_install_names)
error: /Library/Developer/CommandLineTools/usr/bin/install_name_tool: changing install names or rpaths can't be redone for: /Users/User_Name/Tissue-Simulation-Toolkit/lib/muscle3/muscle3/lib/libymmsl.dylib (for architecture arm64) because larger updated load commands do not fit (the program must be relinked, and you may need to use -headerpad or -headerpad_max_install_names)
```
This is related to the long pathname used.

LourensVeen:
Hi, thanks for the report! It's been [showing up](https://github.com/multiscale/muscle3/actions/runs/14941136051/job/41978309149) in the Continuous Integration runs actually, but I haven't had time to look at it yet. Older versions of macOS are fine, so it's something that's changed in Sequoia.

Now that I know that it's actually bothering someone I'll try to give it more priority, but I'm super-busy this weekend and the coming week with other projects, so it'll be after that. Best I can do :sweat:

sbr-shakibi:
Hi!
Thanks for the fast response. Actually, I have figured out the solution and I am currently not bother by it. I wanted to make it public so that no one else would be bothered by it. I will do a pull request (with a solution that works for me) today. I had problems with passing the unit tests (even before my changes), so maybe you can review it when you have time later?

LourensVeen:
Ah, thanks! That looks good, I just need to figure out how to run the OS compatibility actions, because as-is it's only going to be tested on Linux. 

Are you referring to the TST unit tests, or the MUSCLE3 unit tests?

sbr-shakibi:
You're welcome. I was referring to the MUSCLE3 unit tests.

LourensVeen:
[...]

PR #316, opened 2025-05-16
Resolving issue #315 relate to long pathnames

sbr-shakibi:
Adding flags to Makefiles to fix errors mentioned in the issue #315


LourensVeen:
Looks good, as far as I can see this is the right solution. The CI runs are a bit meaningless because we really need the OS compatibility ones, but I have no easy way of starting them here and they're going to run on the timer in a few hours, so I'm just going to merge this and then we'll see. Thanks!

Issue #321, opened 2025-08-13
GoogleTest dependency obsolescent


LourensVeen:
We download GTest 1.12.1 for building the tests, which is giving a CMake deprecation warning. It still works, but it should be updated. 

Related: #317

LourensVeen:
Fix released with 0.9.0, thanks!

PR #323, opened 2025-08-15
Proposed fix for issue #321 (Upgraded GoogleTest)

pranavshridhar:
Hello,

I upgraded the GoogleTest dependency from 1.12.1 to 1.15.2. You shouldn't be getting the CMake warning anymore.
Keep in mind the latest release of GoogleTest available is 1.17.0, which requires a minimum standard of C++17. 1.15.2 is the latest version that still accepts at least C++14. Consider reformatting codebase to fit C++17 standard.

(Proposed fix for #321)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,3 @@ docs/source/examples/*.log
 docs/source/examples/python/performance.sqlite
 
 libmuscle/python/libmuscle/version.py
-
-# Build Artifact from `build cpp`
-libmuscle/cpp/src/libmuscle/version.h-e
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ docs/source/examples/python/performance.sqlite
 libmuscle/python/libmuscle/version.py
 
 libmuscle/cpp/src/libmuscle/version.h-e
+cpp_tests_std*
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,9 @@
 *.egg-info
 *.eggs
 
+.vscode/
+
+
 /dist
 /build
 .tox
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.egg-info
 *.eggs
 
+# Editor environment
 .vscode/
 
 
@@ -25,4 +26,5 @@ docs/source/examples/python/performance.sqlite
 
 libmuscle/python/libmuscle/version.py
 
+# Build Artifact from `build cpp`
 libmuscle/cpp/src/libmuscle/version.h-e
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ docs/source/examples/*.log
 docs/source/examples/python/performance.sqlite
 
 libmuscle/python/libmuscle/version.py
+
+libmuscle/cpp/src/libmuscle/version.h-e
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@
 .pytest_cache
 .mypy_cache
 *.swp
+*.log
 
 htmlcov
 .coverage
@@ -25,4 +26,3 @@ docs/source/examples/python/performance.sqlite
 libmuscle/python/libmuscle/version.py
 
 libmuscle/cpp/src/libmuscle/version.h-e
-cpp_tests_std*
--- a/.gitignore
+++ b/.gitignore
@@ -2,18 +2,12 @@
 *.egg-info
 *.eggs
 
-# Editor environment
-.vscode/
-
-
 /dist
 /build
 .tox
 .cache
 .pytest_cache
 .mypy_cache
-*.swp
-*.log
 
 htmlcov
 .coverage
--- a/libmuscle/cpp/build/Makefile
+++ b/libmuscle/cpp/build/Makefile
@@ -36,7 +36,7 @@ include $(TOOLDIR)/make_available.make
 
 dep_name := googletest
 dep_version_constraint := >= 1.8.1
-dep_version := 1.12.1
+dep_version := 1.17.0
 dep_pkgconfig_name := gtest
 dep_install := 0
 include $(TOOLDIR)/make_available.make
--- a/libmuscle/cpp/build/Makefile
+++ b/libmuscle/cpp/build/Makefile
@@ -36,7 +36,7 @@ include $(TOOLDIR)/make_available.make
 
 dep_name := googletest
 dep_version_constraint := >= 1.8.1
-dep_version := 1.17.0
+dep_version := 1.15.2
 dep_pkgconfig_name := gtest
 dep_install := 0
 include $(TOOLDIR)/make_available.make
--- a/libmuscle/cpp/build/googletest/Makefile
+++ b/libmuscle/cpp/build/googletest/Makefile
@@ -11,15 +11,13 @@ distclean: clean
 	rm -f release-*.tar.gz
 
 
-release-$(googletest_VERSION).tar.gz:
+googletest-$(googletest_VERSION).tar.gz:
 	$(DOWNLOAD) https://github.com/google/googletest/releases/download/v$(googletest_VERSION)/googletest-$(googletest_VERSION).tar.gz
-	mv googletest-$(googletest_VERSION).tar.gz release-$(googletest_VERSION).tar.gz
 
-googletest-release-$(googletest_VERSION): release-$(googletest_VERSION).tar.gz
-	$(TAR) xf release-$(googletest_VERSION).tar.gz
+googletest-release-$(googletest_VERSION): googletest-$(googletest_VERSION).tar.gz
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -2,13 +2,18 @@
 *.egg-info
 *.eggs
 
+# Editor environment
```
LourensVeen: Since different people use different editors, it's generally better to put ignore rules for editor files in a global configuration file. By default on Linux, that's `~/.config/git/ignore`. So please remove this (and yes, that `*.swp` below should also go then).

temp.sh
LourensVeen, line 1: Have you tried the up-arrow in your terminal?

Yes, this needs to go, because we can't have a script for every possible command that someone might want to type in the repository. Also, `temp.sh` isn't a good name for anything, because it's not descriptive of what the script does. Please remove.

.gitignore
```diff
@@ -2,13 +2,18 @@
...
+
+
 /dist
 /build
 .tox
 .cache
 .pytest_cache
 .mypy_cache
 *.swp
+*.log
```
LourensVeen: This can be removed with the `temp.sh` script, because it's not necessary without the script.

libmuscle/cpp/build/Makefile
```diff
@@ -36,7 +36,7 @@ include $(TOOLDIR)/make_available.make
 
 dep_name := googletest
 dep_version_constraint := >= 1.8.1
-dep_version := 1.12.1
+dep_version := 1.15.2
```
LourensVeen, line 39: Great, and good work looking at the C++ version compatibility! Very nice you noticed that and took the right version.

I think we're slowly starting to approach the point where everyone in academia has a C++17 compiler available, but you'd be amazed at how many little compute clusters with way obsolete software universities have stuffed away in closets, still in use :smile:.

libmuscle/cpp/build/googletest/Makefile
```diff
@@ -12,13 +12,14 @@ distclean: clean
 
 
 release-$(googletest_VERSION).tar.gz:
-	$(DOWNLOAD) https://github.com/google/googletest/archive/release-$(googletest_VERSION).tar.gz
+	$(DOWNLOAD) https://github.com/google/googletest/releases/download/v$(googletest_VERSION)/googletest-$(googletest_VERSION).tar.gz
+	mv googletest-$(googletest_VERSION).tar.gz release-$(googletest_VERSION).tar.gz
```
LourensVeen: This rename is superfluous, the rules below can just be changed to use the downloaded name and that makes more sense to me.

libmuscle/cpp/build/googletest/Makefile
```diff
@@ -12,13 +12,14 @@ distclean: clean
...
 
 googletest-release-$(googletest_VERSION): release-$(googletest_VERSION).tar.gz
 	$(TAR) xf release-$(googletest_VERSION).tar.gz
 
-googletest: googletest-release-$(googletest_VERSION)
-	cd googletest-release-$(googletest_VERSION) && mkdir -p build
-	cd googletest-release-$(googletest_VERSION)/build && cmake -DCMAKE_INSTALL_PREFIX=$(CURDIR)/googletest ..
-	cd googletest-release-$(googletest_VERSION)/build && $(MAKE) -j $(NCORES) && make install
-
+googletest: release-$(googletest_VERSION).tar.gz
```
LourensVeen: This rule now does both unpacking and building, why? The rule above does a perfectly good job of unpacking the archive, so why duplicate it?

LourensVeen (changes requested):
Hi Pranav, thank you!

Very nice to see that everything works with the newer version. I did find a few issues with the PR itself which I have added notes on, could you please fix those so we can merge this?

Thanks again!

temp.sh
pranavshridhar: Yep fair enough. That script had a lot of other things I thought I needed but don't.

.gitignore
pranavshridhar: Thank for telling me about this! Honestly this is the first time I am learning about git globalExclude files.

libmuscle/cpp/build/googletest/Makefile
pranavshridhar: My initial logic was changing as little of the build process as possible. Now that we are cleared to use the newer downloaded names, this works great!

pranavshridhar:
Hi, Lourens!

Thank you for your notes. In #324 I have made the improvements discussed in this PR. (edit: they're here)


Edit: completely disregard. I am very new to open source contributions. I closed 324 and the commits are now still in my fork. I can see they're here too.
Edit 2: Sorry for closing. I had a freakout because I was unsure if the commits were visible here or not.

libmuscle/cpp/build/Makefile
pranavshridhar: Glad it's working!

Kind of a dumb question: how difficult would it be to upgrade the compiler in these machines? Even if that’s not possible, newer C++ standards generally maintain strong backward compatibility, so code written in C++14 should still compile with -std=c++17.

LourensVeen: It would entail either installing a newer set of modules, which here in Europe often means fighting EasyBuild, or upgrading the operating system.

It's all possible, but there often just isn't any budget to do it, because the machine was bought long ago from some leftover money with no regard for support, or with a short period of support funded by some project that's already ended.

Then it falls to any resident sysadmins to do these kinds of upgrades, but they have a million other things to do, and so the machine is more or less abandoned, but still in use. (Electricity usually comes from some general budget, so once you have it it doesn't cost anything to keep it, as long as you don't need to maintain it.)

To be clear, this is the long tail I'm talking about, most machines do get updates and do have newer software available, but it's not like we're missing out on all that much by sticking with C++14 for a bit longer. We could replace the `Optional<>` template by `std::optional` and that would save some code (and some compiler warnings), and structured bindings would be nice in a few places, and that's more or less it I think.

libmuscle/cpp/build/googletest/Makefile
LourensVeen: That's an understandable impulse in an unfamiliar code base, but in my opinion it's best to just make the new code as simple as possible. If we want to see what it looked like before then we can just ask git :smile:.

.gitignore
```diff
@@ -20,3 +19,6 @@ docs/source/examples/*.log
 docs/source/examples/python/performance.sqlite
 
 libmuscle/python/libmuscle/version.py
+
+# Build Artifact from `build cpp`
+libmuscle/cpp/src/libmuscle/version.h-e
```
LourensVeen: This looked very weird to me, what's a `.h-e` file, and why would the build system make one?

But I think I've figured it out, you're on a mac, aren't you? The Makefile (in `libmuscle/cpp/build/libmuscle/Makefile`) tries to make `version.h` from `version.h.in` using `sed`, but macos ships with a version of sed that doesn't quite have the same syntax as GNU sed, and if I read the manpage correctly that could explain how this file got made. Also, it would result in `version.h` not being made correctly.

So this file is the result of a bug, and that bug should be fixed rather than the file ignored. Could you remove it here? Then I'll make a PR that fixes the bug and ask you to test it, since I don't have a mac.

LourensVeen (changes requested):
Okay, that's looking good! I've requested one more small change, and then we're going to merge this :smile:

LourensVeen:
Also, no worries about the mess, we all start somewhere. Your second-guess was correct, it's better to continue with the existing PR because that keeps all the conversation together, it's how the system is supposed to be used.

The next step on that road would be to modify the stack of commits where appropriate rather than pushing more on top. For a small PR like this it's fine, because it's easy to review the changes as a whole, but for larger, more complex PRs it's nice to be able to review commit-by-commit, and that gets really confusing if changes are introduced and then reverted again.

LourensVeen:
And we have #326 to hopefully fix the `version.h-e` issue. Could you give that a try?

.gitignore
pranavshridhar: Sure! I'll get on it and track the issue down. We'll continue the conversation in #326 where you mentioned the issue.

pranavshridhar:
> The next step on that road would be to modify the stack of commits where appropriate rather than pushing more on top. For a small PR like this it's fine, because it's easy to review the changes as a whole, but for larger, more complex PRs it's nice to be able to review commit-by-commit, and that gets really confusing if changes are introduced and then reverted again.

Just so I'm clear on this, that would mean editing the commits here on Github, or is there a way to directly edit the commits on my machine?

Edit: Just figured it out, I think

.gitignore
pranavshridhar: It works! We should be good to merge.

LourensVeen: I pushed another commit onto your branch to remove this, since you overlooked it. So that resolves this too.

LourensVeen (approved):
Looks good like this, I'm going to merge this now.

Issue #325, opened 2025-08-20
C++ version.h file not created properly on macos


LourensVeen:
We're using the non-standard argument `-i` to `sed` when we're generating the `version.h` file for libmuscle C++. That works differently on macos, and as a result, on that OS a file named `version.h-e` is created, and `version.h` is incorrect.

The solution is probably to avoid in-place editing, and make sure to use only POSIX-standard `sed`, which both the macos and the GNU version should support.

LourensVeen:
Released with 0.9.0.

PR #326, opened 2025-08-20
Rewrite version.h generation command to work on macos as well

LourensVeen:
I think this addresses #325, but I don't have a mac to test it on. @pranavshridar could you give it a try? There should be no `libmuscle/cpp/src/libmuscle/version.h-e` file now, and `libmuscle/cpp/src/libmuscle/version.h` should contain
```
#pragma once

#define MUSCLE3_VERSION_MAJOR 0
#define MUSCLE3_VERSION_MINOR 8
#define MUSCLE3_VERSION_PATCH 1

#define MUSCLE3_VERSION "0.8.1-dev"
```


pranavshridhar:
Yep! Works! The file `libmuscle/cpp/src/libmuscle/version.h-e` is no longer generated as a result of any make rules. Thanks, Lourens!

Issue #327, opened 2025-08-21
Fix missing mpi.h error on Mac (almost resolved; some questions remain)


pranavshridhar:
Hey, y'all!

I ran into a build problem where `mpi.h` was not being found. I've traced the issue and have a working fix in place.

All I need is a few of questions answered before making a solid PR:
- Between OpenMPI and MPICH, which do we officially support/prefer? 
    - Currently trying to support both in my fix
- Can we reasonably expect `pkg-config` to be available? 
    - This would make it easier to surface the correct `-I[path]` and other needed flags when using MPICH.

Once I get answers to these, I’ll be ready to wrap up the PR and send it over.

LourensVeen:
Hi Pranav!

We support compiling* with any MPI implementation, because MPI is a standard interface and any implementation should work. MPI code is compiled using the `mpicc`, `mpicxx` and `mpifort` compiler wrappers, which will automatically add any necessary flags.

If you're getting an error about `mpi.h`, then there's either something wrong with your MPI setup, or there's a problem in the MUSCLE3 build system. Could you maybe share the build output?

* Since the `mpirun` command doesn't have a standard way of specifying on which subset of resources to start the program, we do support a specific subset for launching instances, which is currently OpenMPI, Intel MPI and launching via SLURM's srun. But that is unrelated to how M3 is compiled.

pranavshridhar:
Current Status: Fix is ready to send right now. Issue #321 needs to be closed first.

---

**TL;DR**: It's something wrong with the build system. I have a fix and will send out a PR once I feel my fix is robust enough. 

# The problem:

## My MPI setup:

Previously, I had none. The code should still compile without that error, since the code block below should prevent that:
```cpp
#ifdef MUSCLE_ENABLE_MPI
#include <mpi.h>
#endif
```

##  The STDERR output
However, I was still getting errors like:
```
In file included from /Users/pranavshridhar/Projects/muscle3/libmuscle/cpp/build/libmuscle/../../src/libmuscle/bindings/libmuscle_fortran_c.cpp:5:
In file included from /Users/pranavshridhar/Projects/muscle3/libmuscle/cpp/build/libmuscle/../../src/libmuscle/libmuscle.hpp:5:
In file included from /Users/pranavshridhar/Projects/muscle3/libmuscle/cpp/build/libmuscle/../../src/libmuscle/instance.hpp:3:
[...]

PR #328, opened 2025-08-22
Improved MPI Compilation behavior on MacOS environments

pranavshridhar:
This PR resolves an issue on macOS (Darwin) where the build fails if MPI is not installed. Previously, the` -DMUSCLE_ENABLE_MPI` macro was being passed unconditionally, causing the compiler to attempt to include `<mpi.h>` even when no MPI implementation was present.

This change ensures that -DMUSCLE_ENABLE_MPI is only defined if MPI is detected on the system, allowing builds to succeed both with and without MPI installed.

The fix has been tested on both macOS and Linux systems, and the problem is resolved on both platforms.

For more context and discussion, see #327.


libmuscle/cpp/build/libmuscle/Makefile
```diff
@@ -31,8 +31,16 @@ endif
 
 tests_deps := libmuscle_d.a version.h
 
+# Checking if we are installing and have MPI libs
+INSTALL_LIBS := $(findstring install,$(MAKECMDGOALS))
+LIBMUSCLE_MPI_SO_PRESENT := $(wildcard libmuscle_mpi.so)
+INSTALLING_WITH_MPI_LIBS := $(and $(INSTALL_LIBS),$(LIBMUSCLE_MPI_SO_PRESENT))
+
 # if we're compiling with MPI, or we are installing and have MPI libs
-ifeq ($(if $(or $(MUSCLE_ENABLE_CPP_MPI),$(and $(findstring install,$(MAKECMDGOALS)),$(wildcard libmuscle_mpi.so))),1,0),1)
```
pranavshridhar, line 35: Refactored the logic in this line to be a bit more cleaner

libmuscle/cpp/build/libmuscle/Makefile
```diff
@@ -157,9 +165,17 @@ LDFLAGS2 += $(shell export PKG_CONFIG_PATH=$(PKG_CONFIG_EXTRA_DIRS):$(PKG_CONFIG
 endif
 
 
+ifdef MUSCLE_ENABLE_CPP_MPI
```
pranavshridhar, line 168: If we are using MPI with C++, we want the wrapper compilers associated with it.

pranavshridhar:
Alright, that wasn't supposed to happen, guess I haven't mastered the art of rebasing yet. But the good news about the error of the unresolved file paths mentioned in a comment in #327? Solved it! I just forgot to include the proper flags, but it's solved!


**Ready to merge, all cpp tests pass!**

pranavshridhar:
Yep, restored now. Sorry about that @LourensVeen

libmuscle/cpp/build/libmuscle/Makefile
LourensVeen: Nice, that's much easier to understand.

libmuscle/cpp/build/libmuscle/Makefile
```diff
@@ -31,8 +31,16 @@ endif
...
 
+# Checking if we are installing and have MPI libs
+INSTALL_LIBS := $(findstring install,$(MAKECMDGOALS))
+LIBMUSCLE_MPI_SO_PRESENT := $(wildcard libmuscle_mpi.so)
+INSTALLING_WITH_MPI_LIBS := $(and $(INSTALL_LIBS),$(LIBMUSCLE_MPI_SO_PRESENT))
+
 # if we're compiling with MPI, or we are installing and have MPI libs
-ifeq ($(if $(or $(MUSCLE_ENABLE_CPP_MPI),$(and $(findstring install,$(MAKECMDGOALS)),$(wildcard libmuscle_mpi.so))),1,0),1)
+USE_MPI := $(or $(MUSCLE_ENABLE_CPP_MPI),$(INSTALLING_WITH_MPI_LIBS))
+MAKE_USE_MPI_FLAG := $(if $(USE_MPI),1,0)
```
LourensVeen: I'd prefer just `USE_MPI_FLAG` here, the `MAKE_` prefix is a bit confusing to me.

libmuscle/cpp/build/libmuscle/Makefile
LourensVeen: Yes, that's the right fix. `CONDITIONAL_CC` isn't very descriptive though, `DEP_CC` would be better and more consistent. I'm nitpicking to be sure, but I seem to be more sensitive to naming than others.

libmuscle/cpp/build/msgpack/Makefile
```diff
@@ -29,7 +29,7 @@ $(MSGPACK_SRC): msgpack-$(msgpack_VERSION).tar.gz
 $(MSGPACK_LIB): $(MSGPACK_SRC)
 	cd msgpack-$(msgpack_VERSION) && rm -rf build && mkdir -p build && cd build && \
 		export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH):$(PKG_CONFIG_EXTRA_DIRS) && \
-		cmake -DCMAKE_INSTALL_PREFIX=$(CURDIR)/msgpack -DMSGPACK_CXX11=ON -DMSGPACK_BUILD_EXAMPLES=OFF -DMSGPACK_BUILD_TESTS=OFF .. && \
+		cmake -DCMAKE_POLICY_VERSION_MINIMUM=3.5 -DCMAKE_INSTALL_PREFIX=$(CURDIR)/msgpack -DMSGPACK_CXX11=ON -DMSGPACK_BUILD_EXAMPLES=OFF -DMSGPACK_BUILD_TESTS=OFF .. && \
```
LourensVeen: This seems to be unrelated to the rest, and should be in a separate PR. I know what issue it solves and this is the right solution, just not here.

LourensVeen (commented):
Nice, this works and it's the right solution. I've left a few comments that I'd like to see fixed, but I'd be happy to do it myself, because they're very minor.

pranavshridhar:
>  I've left a few comments that I'd like to see fixed, but I'd be happy to do it myself, because they're very minor.

No need, I'd be happy to resolve them and send it over.

libmuscle/cpp/build/libmuscle/Makefile
LourensVeen: Thanks!

LourensVeen:
Okay, thank you, I'm going to merge this now.

Looking at it again, I think I'd overlooked the change to look for clang++ first on macos. It kind of makes sense, but then XCode already symlinks g++ to clang++, so that g++ on macos is usually clang++ anyway. If you've used homebrew to install gcc then you get something like g++-14, and you'd probably want to specify exactly which compiler to use anyway.

That said, it's not a bad idea, but if things work then let's leave well enough alone.

LourensVeen (approved):

PR #330, opened 2025-08-28
Leak detection on MacOS uses leaks in the absence of Valgrind

### LINKED ISSUE #329: Proposal: Memory Leak Instrumentation on macOS (leaks as a valgrind alternative)
Hey, y'all!

Much as it annoys me, `valgrind` does not have enough support to run reliably on MacOS systems. However, Apple ships their own memory leak detector called `leaks` that comes with the XCode Command Line Tools. Between that, `valgrind`, and `AddressSanitizer`, `leaks` is the best tool for the task on Darwin systems.

### Proposal: 
* We use `leaks` as the instrumentation for MacOS
* If `valgrind` isn’t found (it won’t be on macOS), fall back to `leaks`.
* If neither `valgrind` nor `leaks` is found, we proceed with the default case that runs tests without them

### Why:
* Right now, `valgrind` is the only memory leak instrumentation used
* `valgrind` cannot be used on modern macOS systems (especially Apple Silicon), and is effectively unsupported.
* Supporting `leaks` makes leak checks universally runnable across platforms.

---

I'm happy to get to work on this right now. I'd just like to hear what y'all think before I proceed.

pranavshridhar:
This PR should resolve #329, plus renaming some variables to keep a more consistent naming convention.

### DIFF
--- a/scripts/gmake/check_tools.make
+++ b/scripts/gmake/check_tools.make
@@ -195,7 +195,6 @@ endif
 
 ifdef MUSCLE_MACOS
     export LEAK_DETECTOR := leaks -atExit --
-    export MallocStackLogging = 1
 endif
 
 ifndef LEAK_DETECTOR
--- a/scripts/gmake/check_tools.make
+++ b/scripts/gmake/check_tools.make
@@ -182,22 +182,27 @@ endif
 
 # Check for valgrind (for testing for memory leaks)
 $(info )
-$(info Looking for valgrind...)
-tool_var := VALGRIND
+$(info Looking for a leak detection tool...)
+tool_var := LEAK_DETECTOR
 include $(TOOLDIR)/check_override.make
 
 tool_command := valgrind
 include $(TOOLDIR)/detect_tool.make
 
-ifeq ($(VALGRIND), valgrind)
-    export VALGRIND := valgrind --leak-check=full --error-exitcode=1
+ifeq ($(LEAK_DETECTOR), valgrind)
+    export LEAK_DETECTOR := valgrind --leak-check=full --error-exitcode=1
 endif
 
-ifndef VALGRIND
+ifdef MUSCLE_MACOS
+    export LEAK_DETECTOR := leaks -atExit --
+    export MallocStackLogging = 1
+endif
+
+ifndef LEAK_DETECTOR
     $(warning - Could not find valgrind, so tests will run without it.)
-    $(warning - To fix this, install valgrind and if necessary set VALGRIND to point to it.)
+    $(warning - To fix this, install valgrind and if necessary set LEAK_DETECTOR to point to it.)
 else
-    $(info - Will check for leaks using $(VALGRIND).)
+    $(info - Will check for leaks using $(LEAK_DETECTOR).)
 endif
 
 # Check number of cores
--- a/libmuscle/cpp/build/libmuscle/tests/Makefile
+++ b/libmuscle/cpp/build/libmuscle/tests/Makefile
@@ -113,7 +113,7 @@ test_dep_lib_paths := $(subst $(space),:,$(foreach DIR,$(DEP_DIRS),$(DIR)/lib))
 
 .PHONY: run_test%
 run_test%: test%
-	export LD_LIBRARY_PATH=$(test_dep_lib_paths):$(LD_LIBRARY_PATH) ; $(VALGRIND) ./$< $(GTEST_OPTIONS)
+	export LD_LIBRARY_PATH=$(test_dep_lib_paths):$(LD_LIBRARY_PATH) ; $(LEAK_DETECTOR) ./$< $(GTEST_OPTIONS)
 
 # This uses setrlimit, which is not compatible with Valgrind, so run without
 .PHONY: run_test_data_memory_use
--- a/libmuscle/fortran/build/libmuscle/tests/Makefile
+++ b/libmuscle/fortran/build/libmuscle/tests/Makefile
@@ -83,5 +83,4 @@ test_dep_lib_paths := $(subst $(space),:,$(foreach DIR,$(DEP_DIRS),$(DIR)/lib))
 
 .PHONY: run_test%
 run_test%: test%
-	export LD_LIBRARY_PATH=$(test_dep_lib_paths):$(LD_LIBRARY_PATH) ; $(VALGRIND) ./$<
-
+	export LD_LIBRARY_PATH=$(test_dep_lib_paths):$(LD_LIBRARY_PATH) ; $(LEAK_DETECTOR) ./$<
--- a/scripts/tests/Makefile
+++ b/scripts/tests/Makefile
@@ -33,4 +33,4 @@ test_use_echolib: libecholib.so echolib.f90 test_use_echolib.f90
 	$(FC) -o $@ $(FFLAGS) $(DEBUGFLAGS) echolib.f90 test_use_echolib.f90 -L. -lecholib
 
 test: test_use_echolib
-	LD_LIBRARY_PATH=$(PWD):$(LD_LIBRARY_PATH) $(VALGRIND) ./test_use_echolib
+	LD_LIBRARY_PATH=$(PWD):$(LD_LIBRARY_PATH) $(LEAK_DETECTOR) ./test_use_echolib


scripts/gmake/check_tools.make
```diff
@@ -182,22 +182,27 @@ endif
...
 
-ifeq ($(VALGRIND), valgrind)
-    export VALGRIND := valgrind --leak-check=full --error-exitcode=1
+ifeq ($(LEAK_DETECTOR), valgrind)
+    export LEAK_DETECTOR := valgrind --leak-check=full --error-exitcode=1
 endif
 
-ifndef VALGRIND
+ifdef MUSCLE_MACOS
+    export LEAK_DETECTOR := leaks -atExit --
```
LourensVeen, line 197: This seems to blindly assume that we'll always have `leaks` available on macos. Is that the case, even if we're e.g. compiling with a compiler from Homebrew or MacPorts?

scripts/gmake/check_tools.make
```diff
@@ -182,22 +182,27 @@ endif
...
+    export MallocStackLogging = 1
```
LourensVeen: Is this going to enable malloc debugging for the compiler and other tools when we're compiling things? Valgrind is currently being applied only to the tests, and I'd like for this to be the same.

LourensVeen (commented):
Looks good, but I have a few questions. Thanks!

scripts/gmake/check_tools.make
pranavshridhar: So far, this will do exactly what valgrind is doing. I have changed the naming of what was previously `$(VALGRIND)` in the other files to be consistent.

scripts/gmake/check_tools.make
pranavshridhar: Yeah, that was a bit of a blunder on my part. I was assuming that the user on Mac would have the Xcode Command Line Tools, which wasn't the right call on my part. Since `leaks` doesn't have a command like `leaks --version`, so the script at `scripts/gmake/check_tools.make` isn't exactly going to work. However, there might be another way to check installation, and I will check it out.

To answer your question: Leaks will (probably) not be installed by default on MacOS devices. However, if it is installed, it will definitely work when we are compiling with any type of C++ Compiler that can run on MacOS systems, which includes the default `clang` that comes with XCode or `g++-15` from Homebrew.

Edit: looks like leaks will always be available on MacOS, which means I can revert to the commit I had before. The prerequisite for both HomeBrew or MacPorts is that XCode Command Line Tools HAS to be installed. As far as I know, Homebrew will even install it manually for you. MacPorts might not, but it still needs it.

`leaks`, `gcc`, and `clang` exist as paths in every MacOS system. While `leaks` is not locked behind Xcode Command Line tools and can therefore run without issue, you really can't use it without a compiler.

That said, I will revert to the commit I had before, since it has been tested and works great.

pranavshridhar:
Standby. It's not working as well as I had hoped.

pranavshridhar:
Now it is

scripts/gmake/check_tools.make
pranavshridhar: Scratch that. We're gonna leave it like this so we don't shoot ourselves in the foot later. Probably shouldn't have changed it, but I am currently experiencing some git issues and this works just as well. I am purposefully leaving out the MallocStackLogging environment variable since it is an intense eyesore when looking at debug logs.

pranavshridhar:
Best to merge #328 and #331 first. I want to explore this one further.

LourensVeen:
#328 and #331 have been merged now. Please let me know if you want me to review again and get this merged.

pranavshridhar:
Thank you Lourens! This branch is technically good to merge, but I wanted to explore what was causing the monstrosity below:
```
g++(31257) MallocStackLogging: could not tag MSL-related memory as no_footprint, so those pages will be included in process footprint - (null)
g++(31257) MallocStackLogging: stack logs being written to /private/tmp/stack-logs.31257.100a64000.g++.9MEsN6.index
g++(31257) MallocStackLogging: recording malloc and VM allocation stacks to disk using standard recorder
clang++(31226) MallocStackLogging: could not tag MSL-related memory as no_footprint, so those pages will be included in process footprint - (null)
clang++(31226) MallocStackLogging: stack logs being written to /private/tmp/stack-logs.31226.10824c000.clang++.sZJfYi.index
clang++(31226) MallocStackLogging: recording malloc and VM allocation stacks to disk using standard recorder
clang++(31226) MallocStackLogging: stack logs deleted from /tmp/stack-logs.31226.1023d4000.xcrun.cIVPNd.index
clang(31194) MallocStackLogging: stack logs deleted from /private/tmp/stack-logs.31194.10c528000.clang.u4JEgP.index
clang++(31141) MallocStackLogging: stack logs deleted from /private/tmp/stack-logs.31141.10ca10000.clang++.wO2kyz.index
mkdir(31266) MallocStackLogging: could not tag MSL-related memory as no_footprint, so those pages will be included in process footprint - (null)
mkdir(31266) MallocStackLogging: stack logs being written to /private/tmp/stack-logs.31266.100dc0000.mkdir.fROZt4.index
mkdir(31266) MallocStackLogging: recording malloc and VM allocation stacks to disk using standard recorder
clang++(31240) MallocStackLogging: could not tag MSL-related memory as no_footprint, so those pages will be included in process footprint - (null)
clang++(31240) MallocStackLogging: stack logs being written to /private/tmp/stack-logs.31240.10c078000.clang++.jFzVm7.index
clang++(31240) MallocStackLogging: recording malloc and VM allocation stacks to disk using standard recorder
clang++(31240) MallocStackLogging: stack logs deleted from /tmp/stack-logs.31240.102308000.xcrun.ftxG3x.index
g++(31257) MallocStackLogging: could not tag MSL-related memory as no_footprint, so those pages will be included in process footprint - (null)
g++(31257) MallocStackLogging: stack logs being written to /private/tmp/stack-logs.31257.100cbc000.g++.CBRzbs.index
g++(31257) MallocStackLogging: recording malloc and VM allocation stacks to disk using standard recorder
g++(31257) MallocStackLogging: stack logs deleted from /tmp/stack-logs.31257.100a64000.g++.9MEsN6.index
clang(31267) MallocStackLogging: could not tag MSL-related memory as no_footprint, so those pages will be included in process footprint - (null)
clang(31267) MallocStackLogging: stack logs being written to /private/tmp/stack-logs.31267.10a3a4000.clang.6OMI1a.index
clang(31267) MallocStackLogging: recording malloc and VM allocation stacks to disk using standard recorder
mkdir(31266) MallocStackLogging: stack logs deleted from /private/tmp/stack-logs.31266.100dc0000.mkdir.fROZt4.index
```
This is only a minuscule fragment of the flood of logs on my console.

# Here's why it's happening
`leaks`, the CLI tool that is used for memory leak instrumentation on MacOS systems, is also commonly used with the `MallocStackLogging=1` environment variable. As you can see, the console noise generated is insane. I had hoped that merging #328 and #331 would reduce the dependency chain here, but unfortunately, that wasn’t the case.

# Current PR-closing resolution
Here's the really fun part: the output from MallocStackLogging doesn’t go to STDOUT or STDERR at all. Instead (Apple being Apple) it is written to files like `/private/tmp/stack-logs.*`, which end up on the console by directly accessing the TTY. The logs can be accessed for observation, but users should not be forced to go through that kind of trouble. What we can do is have the build system not enable MallocStackLogging by default. The good news is that the actual leaks output—the stuff we actually care about—does show up cleanly in STDOUT, as you can see in this test snippet:
```
Process:         test_checkpoint_triggers [92024]
Path:            /Users/USER/*/test_checkpoint_triggers
Load Address:    0x100da4000
Identifier:      test_checkpoint_triggers
Version:         0
Code Type:       ARM64
Platform:        macOS
Parent Process:  leaks [92023]
Target Type:     live task

Date/Time:       2025-09-16 17:53:47.490 -0500
Launch Time:     2025-09-16 17:53:47.123 -0500
OS Version:      macOS 15.6.1 (24G90)
Report Version:  7
Analysis Tool:   /usr/bin/leaks

Physical footprint:         2161K
Physical footprint (peak):  2209K
Idle exit:                  untracked
----

leaks Report Version: 4.0, multi-line stacks
Process 92024: 193 nodes malloced for 14 KB
Process 92024: 0 leaks for 0 total leaked bytes.
```

So for now, I’m going to remove the code that enables MallocStackLogging. If you’re good with it, I’ll also add a note in the documentation advising users to enable it themselves if they want deeper instrumentation. That way we keep the useful output without the log spam.

pranavshridhar:
Good to merge!

LourensVeen (approved):

Issue #317, opened 2025-05-26
MUSCLE3 breaks with msgpack 4.0.0


LourensVeen:
Ubuntu 24.04 ships with msgpack 4.0.0, and MUSCLE3 compiles with it but the all-cpp example test crashes, complaining we received a Data containing the wrong type of data.

The very quick solution would be to change the version compatibility in the build system to be `>=3.2.0,<4` and keep using the old version. Somewhat better would be to test newer versions and fix MUSCLE3 so that it works with those. Then I'm still dreaming of creating my own datapack library instead, and then of course @suvayu wants me to look at Apache Arrow, so we should probably also do that.

Thanks Stéphane for reporting!

pranavshridhar:
The quick solution is exactly what we did in PR #331. For long term stability, the better solution, like you said here, is to test newer versions of MessagePack and fix the MUSCLE3 source code so it works with that, just as we did with the GoogleTest dependency in #323.

The good news is that the latest version of MessagePack 7.0.0 is compatible with C++14 (through some [configuration](https://github.com/msgpack/msgpack-c/blob/cpp_master/README.md#building-and-installing)). However, since M3 is breaking on 4.0.0, so there may be some deprecation that's occurred between 3.2.0 (what we use now), and that. This is how I recommend we proceed:
 - fixing this issue by upgrading to 4.0.0 by default
 - smoothen out the rough edges in the code to be release ready
 - look at upgrading to 7.0.0 for later (optional, but strongly recommended). 

Time permitting, we can even do that last step before the next release.

I'll get to work on this. And also your datapack library if you decide to go through with it 😊.

LourensVeen:
Yes, that looks good, thanks!

By the way, out of curiosity, can I ask you how you found MUSCLE3? It's not exactly world famous :smile:. And why did you decide to contribute? Anyway I'm happy with your help!

Feel free to send me an email at *address redacted*, then we can keep this issue about MsgPack.

pranavshridhar:
Just sent you an email! As for this issue, I seem to have found a cause for this, but I am doing some dependency analysis to get a good understanding of how to implement an informed fix. It'll take me maybe a week at most, but I'll update you on my progress by editing this message.

LourensVeen:
Sure, no hurry! The MsgPack API isn't great, and isn't greatly documented, and also the Data and DataConstRef stuff is a bit hairy, so I'd expect it to take a bit of time to understand what's going on.

 Thanks!

[...]

PR #331, opened 2025-09-10
Updated MessagePack's Minimum Version Policy in its Makefile

pranavshridhar:
In #328, I fixed a deprecation error related to GNU Make's version policy with MessagePack. Since that change warranted its own PR, I’m opening this one to address it separately.


LourensVeen (approved):
Yes, I've seen this before with other software. The bigger issue is that we're using a rather old msgpack (see #317), and that msgpack-c++ isn't great. But this will at least let us build with newer CMake, which is nice. Thanks!

pranavshridhar:
No problem! I will look more into the msgpack problems here, actually, and mention my findings.

PR #333, opened 2025-10-24
Upgrade from messagepack 3.2.0 to 7.0.0

pranavshridhar:
Hey, y'all!

Finally upgraded MessagePack to the latest dependency of `7.0.0`. This should solve #317, but it's not quite good to merge yet. I am sure that there are better ways to do this.

## Problem: `Data`/`DataConstRef` receives the wrong type
The error faced by the user stated in #317 is a runtime type error, and after a lot of digging I know why: MessagePack 4.0.0, for the sake of minimizing message sizes, packs floating point types with no fractional component as the smallest integer type it can (e.g. 2.0 becomes 2, while 2.3 stays 2.3). I have tried like hell to debug MUSCLE3 to stop this (there is a way, but it involves more time than it should take), but without success.

## Solution: Upgrade to the most recent version
At least this way, we're not locking ourselves in time with a version that doesn't work, but there are changes to how the build system deals with MessagePack.

### MessagePack C/C++ has split
One of the breaking changes in 4.0.0 is the split between the C and the C++ libraries, which are `msgpack-c` and `msgpack-cxx` respectively. Thus, the GitHub URL had to be changed to match this. 

### `msgpack-cxx` does not use `pkgconf`
There is no such thing as a `.pc` file or a `.pc.in` within the `msgpack-cxx` library, so there is no way to use `pkg-config` to get any information about it. Only the C library has it. I was a bit surprised that any compilation was successful under 4.0.0. However, `libmsgpack-dev` lists a `libmsgpack-cxx-dev` as a dependency. I have tried to install this to see if that would be usable with `pkg-config`, but I was unsuccessful, which tells us that the C++ library cannot be used with it at all. The C++ library is also header-only, which might explain the lack of use of the `pkgconf` template files.

### Future detection and installation: 
There is a way that is described in the msgpack-cxx README which is described as the [canonical cmake way](https://github.com/msgpack/msgpack-c/tree/cpp_master?tab=readme-ov-file#usage), but I'm guessing that's not something we want to do.
To that end, the changes in the build system locally download and extract msgpack regardless if the current system has it or not. 4.0.0 has that problematic type compression behavior mentioned earlier, and it's the same version that's the latest available on apt, so this is probably for the better. 

---
There are a lot of lines here that I commented out rather than deleted out of anticipation that we might need to keep it on hand. However, after all issues have been ironed out in this PR, I want to get rid of these lines before merging.

What needs to be done now:
- [x] The minimum required version for MessagePack HAS to be 7.0.0 in the build system. Anything else will invite the same behavior again.
- [ ] Every instance of the string `msgack` in the build system HAS to be changed to `msgpack-cxx`, if it isn't already.
- [ ] An MPI issue in the test suite similar to the one solved in #328 is showing up again. Not to be solved here, but it needs its own fix.
- [ ] **Suggested by @LourensVeen:** We are going to install msgpack, so might as well do away with detection and skip to installing.
- [ ] **Suggested by @LourensVeen:** The documentation should reflect this change as well.
- [x] **Suggested by @LourensVeen:** (see his comment below) Remove msgpack as a dependency listed in `libmuscle.pc` and `libmuscle_mpi.pc` to fix `make test_examples`.

### DIFF
--- a/libmuscle/cpp/build/Makefile
+++ b/libmuscle/cpp/build/Makefile
@@ -28,9 +28,9 @@ distclean:
 # Find dependencies or set up build
 
 dep_name := msgpack
-dep_version_constraint := >= 3.2.0
-dep_version := 3.2.0
-dep_pkgconfig_name := msgpack
+dep_version_constraint := >= 4.0.0
+dep_version := 7.0.0
+dep_pkgconfig_name := msgpack-cxx
 dep_install := 1
 include $(TOOLDIR)/make_available.make
 
@@ -46,7 +46,7 @@ include $(TOOLDIR)/make_available.make
 DEP_DIRS += $(CURDIR)/msgpack/msgpack
 export DEP_DIRS
 
-PKG_CONFIG_EXTRA_DIRS := $(PKG_CONFIG_EXTRA_DIRS):$(CURDIR)/msgpack/msgpack/lib/pkgconfig
+PKG_CONFIG_EXTRA_DIRS := $(PKG_CONFIG_EXTRA_DIRS):$(CURDIR)/msgpack/lib/pkgconfig
 export PKG_CONFIG_EXTRA_DIRS
 
 
--- a/libmuscle/cpp/build/Makefile
+++ b/libmuscle/cpp/build/Makefile
@@ -28,7 +28,7 @@ distclean:
 # Find dependencies or set up build
 
 dep_name := msgpack
-dep_version_constraint := >= 4.0.0
+dep_version_constraint := >= 7.0.0
 dep_version := 7.0.0
 dep_pkgconfig_name := msgpack-cxx
 dep_install := 1
--- a/libmuscle/cpp/build/libmuscle/Makefile
+++ b/libmuscle/cpp/build/libmuscle/Makefile
@@ -141,7 +141,7 @@ install: all $(installed_headers) $(installed_pkg_config_files)
 	# install_name_tool -change @rpath/libmsgpackc.2.dylib @loader_path/libmsgpackc.2.dylib $(PREFIX)/lib/libmuscle.dylib
 	# -install_name_tool -change @rpath/libmsgpackc.2.dylib @loader_path/libmsgpackc.2.dylib $(PREFIX)/lib/libmuscle_d.dylib
 	# -install_name_tool -change @rpath/libmsgpackc.2.dylib @loader_path/libmsgpackc.2.dylib $(PREFIX)/lib/libmuscle_mpi.dylib
-	# -install_name_tool -change @rpath/libmsgpackc.2.dylib @loader_path/libmsgpackc.2.dylib $(PREFIX)/lib/libmuscle_mpi_d.dylib 
+	# -install_name_tool -change @rpath/libmsgpackc.2.dylib @loader_path/libmsgpackc.2.dylib $(PREFIX)/lib/libmuscle_mpi_d.dylib
 	@mkdir -p $(PREFIX)/bin
 	sed -e 's@%PREFIX%@$(PREFIX)@g' muscle3.env.in >$(PREFIX)/bin/muscle3.env
 	for pc in $(PREFIX)/lib/pkgconfig/* ; do sed -i -e 's@^prefix=.*$$@prefix=$(PREFIX)@' $$pc ; done
@@ -289,9 +289,9 @@ libmuscle.pc:
 	@echo 'URL: https://muscle3.readthedocs.io' >>$@
 	@echo 'Version: $(muscle_version)' >>$@
 	@echo 'Requires: ymmsl = $(muscle_version)' >>$@
-	@echo 'Requires.private: msgpack >= 3.1.0' >>$@
-	@echo 'Cflags: -I$${includedir} -pthread' >>$@
-	@echo 'Cflags.private: -pthread' >>$@
+	@echo 'Requires.private:' >>$@
+	@echo 'Cflags: -I$${includedir} -pthread -DMSGPACK_NO_BOOST' >>$@
+	@echo 'Cflags.private: -pthread -DMSGPACK_NO_BOOST' >>$@
 	@echo 'Libs: -L$${libdir} -lmuscle' >>$@
 	@echo 'Libs.private: -pthread' >>$@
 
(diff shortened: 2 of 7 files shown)


libmuscle/cpp/build/libmuscle/Makefile
```diff
@@ -136,10 +136,12 @@ install: all $(installed_headers) $(installed_pkg_config_files)
 	-install_name_tool -change libymmsl.dylib @loader_path/libymmsl.dylib $(PREFIX)/lib/libmuscle_d.dylib
 	-install_name_tool -change libymmsl.dylib @loader_path/libymmsl.dylib $(PREFIX)/lib/libmuscle_mpi.dylib
 	-install_name_tool -change libymmsl.dylib @loader_path/libymmsl.dylib $(PREFIX)/lib/libmuscle_mpi_d.dylib
-	install_name_tool -change @rpath/libmsgpackc.2.dylib @loader_path/libmsgpackc.2.dylib $(PREFIX)/lib/libmuscle.dylib
-	-install_name_tool -change @rpath/libmsgpackc.2.dylib @loader_path/libmsgpackc.2.dylib $(PREFIX)/lib/libmuscle_d.dylib
-	-install_name_tool -change @rpath/libmsgpackc.2.dylib @loader_path/libmsgpackc.2.dylib $(PREFIX)/lib/libmuscle_mpi.dylib
-	-install_name_tool -change @rpath/libmsgpackc.2.dylib @loader_path/libmsgpackc.2.dylib $(PREFIX)/lib/libmuscle_mpi_d.dylib
+
+	# Not needed for MessagePack 7.0.0
```
LourensVeen: Is this because there's no library anymore and it's header-only? And if it's not needed anymore, then we can just delete it, no point in leaving dead code around.

libmuscle/cpp/build/libmuscle/Makefile
```diff
@@ -152,11 +154,14 @@ ifeq "$(filter $(MAKECMDGOALS),$(cleantargets))" ""
 
 # Dependencies
 $(info pcextra: $(PKG_CONFIG_EXTRA_DIRS))
-CXXFLAGS += $(shell export PKG_CONFIG_PATH=$(PKG_CONFIG_EXTRA_DIRS):$(PKG_CONFIG_PATH) ; pkg-config --cflags msgpack)
+# MSGPACK_FLAGS := $(shell export PKG_CONFIG_PATH=$(PKG_CONFIG_EXTRA_DIRS):$(PKG_CONFIG_PATH) ; pkg-config --cflags msgpack-cxx)
```
LourensVeen: Does this mean we'll now always build messagepack locally? If not, how do we detect the library? Or are we assuming that it'll be available in a system path?

libmuscle/cpp/build/msgpack/Makefile
```diff
@@ -17,28 +17,29 @@ distclean: clean
...
+$(MSGPACK_SRC): msgpack-cxx-$(msgpack_VERSION).tar.gz
+	$(TAR) xf msgpack-cxx-$(msgpack_VERSION).tar.gz -m
 
-$(MSGPACK_LIB): $(MSGPACK_SRC)
-	cd msgpack-$(msgpack_VERSION) && rm -rf build && mkdir -p build && cd build && \
+$(MSGPACK_INCLUDE_DIR): $(MSGPACK_SRC)
+	cd msgpack-cxx-$(msgpack_VERSION) && rm -rf build && mkdir -p build && cd build && \
 		export PKG_CONFIG_PATH=$(PKG_CONFIG_PATH):$(PKG_CONFIG_EXTRA_DIRS) && \
-		cmake -DCMAKE_POLICY_VERSION_MINIMUM=3.5 -DCMAKE_INSTALL_PREFIX=$(CURDIR)/msgpack -DMSGPACK_CXX11=ON -DMSGPACK_BUILD_EXAMPLES=OFF -DMSGPACK_BUILD_TESTS=OFF .. && \
+		cmake -DCMAKE_POLICY_VERSION_MINIMUM=3.5 -DMSGPACK_USE_BOOST=OFF -DCMAKE_INSTALL_PREFIX=$(CURDIR)/msgpack -DMSGPACK_CXX11=ON -DMSGPACK_BUILD_EXAMPLES=OFF -DMSGPACK_BUILD_TESTS=OFF .. && \
```
LourensVeen, line 29: Is the CMAKE_POLICY_VERSION_MINIMUM still needed when building 7.0.0.?

libmuscle/cpp/build/Makefile
```diff
@@ -28,9 +28,9 @@ distclean:
 # Find dependencies or set up build
 
 dep_name := msgpack
-dep_version_constraint := >= 3.2.0
-dep_version := 3.2.0
-dep_pkgconfig_name := msgpack
+dep_version_constraint := >= 4.0.0
```
LourensVeen: I think you mentioned that MUSCLE3 breaks with version 4, because of the int/double type confusion issue (what were they thinking?) So should this perhaps be 7.0.0 as a minimum?

LourensVeen: Or actually, if we're always going to build locally (which makes sense if it's header only and they break things regularly), maybe we should remove the detection completely?

LourensVeen (commented):
The number of lines changed in this PR belies the amount of work that went into it. Very nice, and thank you!

LourensVeen:
I've seen this issue of CMake-built packages insisting that everyone else use CMake too if they want to use the library. It's worrying. On the other hand, if this is just about standardising on CMAKE_PREFIX_PATH then it's easy enough to add support to other build systems I suppose. Well, maybe not relevant here if we'll just always build locally.

I've found two more issues while testing this locally:

- [ ] The documentation still lists MsgPack 3.2.0 or later as a dependency, that should either have its version updated, or perhaps we should update it to point out that it will always be downloaded automatically?
- [ ] `make test_examples` fails, because MUSCLE3's own pkg-config files still list msgpack as a dependency. See `libmuscle/cpp/build/libmuscle/libmuscle.pc` and `libmuscle_mpi.pc`. I think that can be removed if it's header-only.

libmuscle/cpp/build/Makefile
pranavshridhar: Removing detection and installing isn't a bad idea and does save some cycles on an unnecessary step. I'll add it to the list!

libmuscle/cpp/build/msgpack/Makefile
pranavshridhar: CMake does NOT like when I do that, so my recommendation is to keep it for now.

libmuscle/cpp/build/Makefile
pranavshridhar: Come to think of it, this isn't an immediate issue and can be fixed later. Detecting and installing with what we have is the easiest thing to do right now.

LourensVeen: Fair enough, the build system needs an overhaul anyway, so I'll address it when I get around to that.

pranavshridhar:
I am genuinely surprised that this is passing as many test cases as it is now. I am facing issues with Ubuntu regarding libmuscle's pc files, but the good news is the Mac build works!

Issue #57, opened 2020-05-13
Add caching proxy


LourensVeen:
In some cases, a macromodel may repeatedly call a micromodel with the same input (e.g. because a boundary condition is passed that doesn't change at every timestep). In that case, you may want to optimise the micromodel by storing the last input and output, and when a new input is received compare it to the last one and if they're identical, return the same output. (This works only for deterministic models of course.)

But why change the micromodel? This is a perfect job for a standardised proxy, which should come with MUSCLE 3.

v1kko:
I would like to have a look at this

LourensVeen:
Released in 0.9.0, albeit as an example rather than a standard component. But we'll take that, and maybe revisit this in the future when we have the new Python-based plug-in support.

PR #335, opened 2025-11-12
Add an example cache written in python

v1kko:
The cache caches the last 128 responses, and does not run the micro model if a cached response is available

- [x] set number of cached responses via settings
- [x] create a test case in which the cache is actually hit
- [x] Update the documentation

~~- [ ] (Optional) implement in cpp and fortran~~


v1kko:
@LourensVeen I think this pull request is ready for review
Closes #57 ( I hoped this would work, but I seem to not be able to attach an issue (or reviewer) to this pull request)

v1kko:
@LourensVeen Trailing whitespaces are removed

LourensVeen (approved):
My only complaint is that I have nothing to complain about. Now what do I do? I suppose I could make a fuss over the a, b, c single-letter variable names, but then I put a U into the other examples as well. So I'm just going to approve and merge this, as the CI failure is unrelated. Thanks!

Issue #312, opened 2025-02-25
Way to setup a monitoring command for the duration of a MUSCLE3 run?


DavidPCoster:
I would like to run a program that starts at the beginning of a muscle3 run, and finishes at the end.

(The starting point was to follow the load average on the node running the code.)

I have not been able to get it to work cleanly since the code typically finishes early when the first send to the code has been received and the (too clever) environment decides no future sends are possible.

Is such a situation easy to set up?

LourensVeen:
I think the easiest way to do that would be to use `srun` in your batch script, outside of MUSCLE3. The command
```
srun --nodes=${SLURM_JOB_NUM_NODES} --ntasks=${SLURM_JOB_NUM_NODES} --ntasks-per-node=1 <command>
```
should start one copy of `<command>` on each node. On SLURM > 20.02 you'll want to add `--overlap` to the `srun` command line to keep SLURM from declaring the nodes occupied, which would keep MUSCLE3 from launching its agents and the model instances.

Of course, MUSCLE3 already records some performance data, and I think it makes sense to extend that with e.g. CPU and memory usage. And the next release of MUSCLE3 has its own launcher, which starts an agent on each node...

So, if libmuscle were to determine the host it's running on, the PID, and the MPI rank where applicable, and would send that information to the manager, then the manager could send that information to the agent on that node, and ask it to monitor that PID, sending the data to the manager for adding to the profile database. Then you'd get things like CPU and memory usage, but connected to the instance id and MPI rank and ready to combine with the MUSCLE3-level data.

I could work out that TODO-list in a bit more detail if you or anyone is willing to have a go at contributing this...

PR #336, opened 2025-11-28
Better monitoring

v1kko:
This PR implements a way to get the node agents (if applicable) to monitor each instance that registers itself in the instance registry

The monitoring is quite basic and can be extended, it is logged with a DEBUG level to the main log.

- [x] Implement more testing for the profiling
- [x] **Important** Send the monitoring data less frequently than every 0.1s, find out how we want to do this
  - It turns out that psutil has to do cpu profiling over a certain period, this would block the main thread, so I created another thread that takes approx 1 second to do this. Then this result is sent back only when it is ready, thus about once every 1.1 seconds.
- [x] Do we support MPI out-of-the-box, or do we add support for that later?
   - something for later

After this is merged, probably there should also be a way to easily extract this info from the database


v1kko:
Closes #312

libmuscle/cpp/src/libmuscle/mmp_client.cpp
```diff
@@ -191,7 +191,8 @@ void MMPClient::register_instance(
     auto request = Data::list(
             static_cast<int>(RequestType::register_instance),
             static_cast<std::string>(instance_id_), encoded_locs,
-            encoded_ports, MUSCLE3_VERSION);
+            encoded_ports, get_process_id(), get_hostname(),
```
LourensVeen: I would much prefer if these were passed in, because as is it extends the scope of this class from implementing the MMP protocol to doing that and also inspecting the environment.

libmuscle/python/libmuscle/manager/mmp_server.py
```diff
@@ -146,13 +146,15 @@ def close(self) -> None:
 
     def _register_instance(
             self, instance_id: str, locations: List[str],
-            ports: List[List[str]], version: str = '') -> Any:
+            ports: List[List[str]], pid: int, hostname: str, version: str = '') -> Any:
```
LourensVeen, line 152: Instances that use MPI will have a hostname/pid for each rank. How would that be passed here?

libmuscle/python/libmuscle/native_instantiator/test/test_process_manager.py
```diff
@@ -72,7 +72,7 @@ def test_exit_code(lpm, tmp_path):
 def test_multiple(lpm, tmp_path):
     for i in range(3):
         lpm.start(
-                f'test_{i}', tmp_path, ['bash', '-c', 'sleep 1'], {},
+                f'test_{i}', tmp_path, ['bash', '-c', '/usr/bin/env sleep 1'], {},
```
LourensVeen: Can you clarify how you created a situation in which `/usr/bin` wasn't in the path? `env` (and `ls` and `sleep`) are all POSIX, so I think it's fair to assume that they're available. But maybe I'm missing something?

libmuscle/python/libmuscle/manager/profile_store.py
```diff
@@ -139,7 +139,7 @@ def to_tuple(e: ProfileEvent) -> Record:
                     instance_oid, e.event_type.value, e.start_time.nanoseconds,
                     e.stop_time.nanoseconds, port_name, port_operator,
                     e.port_length, e.slot, e.message_number, e.message_size,
-                    e.message_timestamp)
+                    e.message_timestamp, e.cpu_percent, e.memory_usage)
```
LourensVeen: I'm guessing that both cpu_percent and memory_usage are averages? Or is memory_usage a maximum? That would probably actually be more useful.

libmuscle/python/libmuscle/manager/test/test_profile_database.py
```diff
@@ -137,3 +140,15 @@ def test_time_taken(db_file):
...
         assert 200.0 == db.time_taken(etype='DEREGISTER', aggregate='sum')
         assert 600.0 == db.time_taken(etype='SEND', aggregate='sum')
+
+def test_resource_usage_event(db_file):
+    with ProfileDatabase(db_file) as db:
+        cur = db._get_cursor()
+        cur.execute(f"SELECT * FROM events WHERE event_type_oid = '{ProfileEventType.RESOURCE_USAGE.value}'")
+        rows = cur.fetchall()
+        assert len(rows) == 1
+        # Check columns (cpu_percent is 2nd to last, memory_usage is last)
```
LourensVeen: Until we add more I guess?

libmuscle/python/libmuscle/manager/mmp_server.py
```diff
@@ -79,7 +81,8 @@ def __init__(
             topology_store: TopologyStore,
             snapshot_registry: SnapshotRegistry,
             deadlock_detector: DeadlockDetector,
-            run_dir: Optional[RunDir]
+            run_dir: Optional[RunDir],
+            instance_manager: Optional[InstanceManager] = None,
```
LourensVeen, line 84: This shouldn't be optional, there's always an InstanceManager, so we can always pass it. I really want to have the runtime structure as inflexible (and therefore comprehensible) as possible.

libmuscle/python/libmuscle/manager/mmp_server.py
```diff
@@ -173,6 +177,10 @@ def _register_instance(
                     ' same version of libmuscle.']
         port_objs = [decode_port(p) for p in ports]
         instance = Reference(instance_id)
+
+        if self._instance_manager:
+            self._instance_manager._requests_out.put(MonitorRequest(instance_id, hostname, pid))
```
LourensVeen: This accesses a private variable, which may be allowed by Python but not by me, at least in this context. There should be an `InstanceManager.monitor_process()` function that does this.

libmuscle/python/libmuscle/manager/mmp_server.py
```diff
@@ -418,7 +426,8 @@ def __init__(
             topology_store: TopologyStore,
             snapshot_registry: SnapshotRegistry,
             deadlock_detector: DeadlockDetector,
-            run_dir: Optional[RunDir]
+            run_dir: Optional[RunDir],
+            instance_manager: Optional[InstanceManager]
```
LourensVeen, line 429: Same here about optionality, and before run_dir with the other system components makes more sense.

libmuscle/python/libmuscle/mcp/protocol.py
```diff
@@ -34,6 +34,7 @@ class RequestType(Enum):
     REPORT_RESOURCES = 41
     GET_COMMAND = 42
     REPORT_RESULT = 43
+    MONITOR_USAGE = 44
```
LourensVeen: But it's not telling the manager to monitor anything, is it? So this should be called `REPORT_USAGE`?

libmuscle/python/libmuscle/native_instantiator/agent/map_client.py
```diff
@@ -44,6 +45,32 @@ def report_resources(self, resources: OnNodeResources) -> None:
                 resources.node_name, {'cpu': enc_cpu_resources}]
         self._call_agent_manager(request)
 
+    def monitor_usage(self, pids: List[Tuple[str, int]], logger) -> None:
```
LourensVeen: Huh? This is a protocol class for the MAP protocol. It's got nothing to do with monitoring resources. Why is this here? I don't get it. And why are we passing a logger object (and where's the type annotation?), instead of just giving this file its own like everywhere else?

libmuscle/python/libmuscle/native_instantiator/agent/map_client.py
```diff
@@ -44,6 +45,32 @@ def report_resources(self, resources: OnNodeResources) -> None:
...
+                logger.debug(f'PID {pid}: Process not found')
+
+        if len(usage) < 1:
+            """ Nothing to monitor """
+            return
+
+        request = [
+                RequestType.MONITOR_USAGE.value,
+                self._node_name, usage]
+        self._call_agent_manager(request)
```
LourensVeen: I think you mentioned that you wanted to use the Profiler and report to the manager in the same way, but now it seems that the MAP has a new endpoint for reporting events. Is the intention to change it still?

libmuscle/python/libmuscle/manager/instance_manager.py
```diff
@@ -172,23 +179,33 @@ def cancel_all() -> None:
         results: List[Process] = list()
 
         while self._num_running > 0:
-            result = self._results_in.get()
+            try:
```
LourensVeen: This really makes me want to give the agent an MMPClient and report the profiling data through that. Would save a bunch of code. (For the record, the background process here is needed because we need to fork to start the agents, and there are threads in various places. We're careful to start this background process before spawning any threads, which avoids angering the POSIX gods.)

libmuscle/python/libmuscle/manager/profile_store.py
```diff
@@ -102,6 +102,20 @@ def store_resources(self, resources: Dict[Reference, ResourceAssignment]) -> Non
         cur.execute("COMMIT")
         cur.close()
 
+    def add_event(
```
LourensVeen, line 105: Why a separate function? Can't we just call `add_events(instance_id, [event])` directly instead?

libmuscle/python/libmuscle/native_instantiator/agent_manager.py
```diff
@@ -81,10 +85,32 @@ def start(
...
+        
+        Args:
+            instance: Name of the instance to monitor
+            hostname: Hostname of the node to monitor
+            pid: Process id to monitor
+        """
         self._server.deposit_command(hostname, AddMonitorCommand(instance, hostname, pid))
 
+    def monitor_usage(self, usage: Dict[int, Tuple[float, int]]) -> None:
+        """Send back a monitor event to the profile store via the queue to the instance manager
```
LourensVeen: So far, the AgentManager didn't know anything about the existence of an InstanceManager. That's good, because InstanceManager is in a higher layer, and we don't want upward dependencies.

It's also not needed: for finished processes we have `report_result()` store the received data locally, and then have the NativeInstantiator call `get_finished()` to pick up those results. If we did the same thing for the profiling data, then there would be one consistent way of transporting information from the agent up, and we wouldn't need to add a dependency here on multiprocessing, or knowledge of the fact that we're running in a background thread and therefore need a queue.

libmuscle/python/libmuscle/native_instantiator/native_instantiator.py
```diff
@@ -50,7 +52,7 @@ def run(self) -> None:
             logs_dir = self._run_dir / 'logs'
             logs_dir.mkdir(exist_ok=True)
 
-            self._agent_manager = AgentManager(logs_dir)
+            self._agent_manager = AgentManager(logs_dir, self._profile_events_out)
```
LourensVeen: Instead of passing this here, and introducing a direct link between the layer above us and the one below us, it would be better to add a `self._report_profiling_data()` to `self._report_failed_processes()` and `self._report_finished_processes()` below. See other comment.

libmuscle/python/libmuscle/manager/instance_manager.py
LourensVeen: On second thought, that's also not nice because now the agents end up knowing about the manager, where currently they live in their own little microcosm in which there's only the AgentManager. Separation is good. And we're really just adding another thing that gets reported by the agents, besides exit codes. If it's implemented in the same way (see other comments) then that's not so bad.

This loop spinloops now though, and will eat a core. Using `get()` with a small timeout would solve that. And probably this whole loop should be moved to a separate function because this thing is too long already. Actually, this is a perfect example of a function that contains subheadings, and those almost always can and should be split up accordingly. Bad Lourens...

LourensVeen (commented):
Okay, I've left a bunch of comments, but I also still have some second thoughts whether it's good to merge this with the profiling events. It may be better to have a separate table (instance, rank, hostname, pid) in the ProfileStore, and another one with (hostname, pid, time_start, time_stop, cpu, mem_max), and then you can join that together and/or with the other data in all sorts of creative ways when analysing the results.

Let me sleep on that, and maybe discuss tomorrow?

LourensVeen:
Oh, and could you merge develop into your branch? I've fixed the CI, so then we can see what it says.

libmuscle/cpp/src/libmuscle/mmp_client.cpp
v1kko: Fixed in commit b54c0e3, it is now passed from instance.cpp

libmuscle/python/libmuscle/manager/mmp_server.py
v1kko: Only the process calling "register_instance" will be monitored, if we want to also monitor all the other MPI processes, then I think that the following needs to be done:

- The  MPI Process that calls register_instance must provide all PIDS and hostnames
- We need to call this method for each provided PID/Hostname Pair

I think the backend is flexible enough to have Many PIDs for a single Instance ID

libmuscle/python/libmuscle/native_instantiator/test/test_process_manager.py
v1kko: This is solved in d1d1a3

libmuscle/python/libmuscle/manager/test/test_profile_database.py
v1kko: Fixed by adding another table

v1kko: and fixed by counting forward in 154f29b

libmuscle/python/libmuscle/manager/mmp_server.py
v1kko: As discussed: the InstanceManager is Optional, as muscle can run without it. If you want to make the InstanceManager mandatory, I think that should be another PR

libmuscle/python/libmuscle/manager/mmp_server.py
v1kko: Fixed in 1366ede

libmuscle/python/libmuscle/native_instantiator/agent/map_client.py
v1kko: Fixed by moving it to a seperate MLPClient

libmuscle/python/libmuscle/manager/instance_manager.py
v1kko: Fixed (well at least the spinlock) by moving to a separate MLPServer

libmuscle/python/libmuscle/manager/profile_store.py
v1kko: I made it because a single agent might submit for various instances, which with `add_events` means you have to pre-sort, or submit with arrays of length 1. but either way is fine I think.

libmuscle/python/libmuscle/native_instantiator/agent_manager.py
v1kko: Fixed with the new MLPclient

libmuscle/python/libmuscle/native_instantiator/native_instantiator.py
v1kko: Fixed by using another path via the MLPServer

PR #340, opened 2026-02-27
add a default value to get_settings

IrisvdWerf:
To do:

- [x] C++ implementation
- [x] C/Fortran implementation
- [x] Update user docs


libmuscle/python/libmuscle/test/test_instance.py
```diff
@@ -312,7 +312,55 @@ def test_get_setting(instance, settings_manager):
...
+    result = instance.get_setting('test3', default=3.14)
+    assert result == 3.14
+
+    settings_manager.get_setting.return_value = True
+    result = instance.get_setting('test4', default=True)
+    assert result is True
+
+    settings_manager.get_setting.return_value = [1, 2, 3]
+    result = instance.get_setting('test5', default=[1, 2, 3])
+    assert result == [1, 2, 3]
```
maarten-ic: This is only testing that `Mock.return_value` is working as it should, so can be removed I think?

libmuscle/python/libmuscle/test/test_instance.py
```diff
@@ -312,7 +312,55 @@ def test_get_setting(instance, settings_manager):
...
+
+
+def test_get_setting_without_default_raises_keyerror(instance, settings_manager):
+    settings_manager.get_setting.side_effect = KeyError('Setting not found')
+    with pytest.raises(KeyError):
+        instance.get_setting('nonexistent')
```
maarten-ic: Can also be removed IMO

libmuscle/python/libmuscle/instance.py
maarten-ic, line 305: We need to update these as well to make the static type checkers happy:

```suggestion
    @overload
    def get_setting(self, name: str, typ: Literal['str'], *, default: Optional[str] = None) -> str:
        ...
```

And similar for the other ones that are following

libmuscle/python/libmuscle/settings_manager.py
```diff
@@ -142,12 +149,14 @@ def get_setting(self, instance: Reference, setting_name: Reference,
                 value = self.base[name]
                 break
         else:
-            raise KeyError(('Value for setting "{}" was not set.'.format(
-                setting_name)))
+            if default is not None:
+                return default
+            raise KeyError(('Value for setting "{}" was not set.'.format(setting_name)))
```
maarten-ic: While we're touching this, we could make this slightly simpler with an f-string:


```suggestion
            raise KeyError(f'Value for setting "{setting_name}" was not set.')
```

libmuscle/python/libmuscle/settings_manager.py
```diff
@@ -142,12 +149,14 @@ def get_setting(self, instance: Reference, setting_name: Reference,
...
 
         if typ is not None:
             if not has_setting_type(value, typ):
-                raise TypeError('Value for setting "{}" is of type {},'
-                                ' where {} was expected.'.format(
-                                    name, type(value), typ))
+                raise TypeError(
+                        'Value for setting "{}" is of type {},'
+                        ' where {} was expected.'.format(
+                            name, type(value), typ))
```
maarten-ic: I'd undo this change, or update it to use an f-string here as well :)

maarten-ic (commented):
Thanks! Couple of comments that you can have a look at, let me know if you have any questions!

libmuscle/cpp/src/libmuscle/instance.hpp
maarten-ic, line 313: Can you implement a default for `get_setting_as` as well?

libmuscle/cpp/src/libmuscle/tests/mocks/mock_settings_manager.hpp
```diff
@@ -11,6 +12,38 @@
...
+            ::ymmsl::Reference const & instance,
+            ::ymmsl::Reference const & setting_name,
+            Optional<::ymmsl::SettingValue> const & default_value = {}) const {
+        return BaseMockFun::operator()(instance, setting_name, default_value);
+    }
+
+    // Overload called_with to support 2 arguments (with default third argument)
+    template <typename A1, typename A2>
+    bool called_with(A1 arg1, A2 arg2) const {
+        return BaseMockFun::called_with(arg1, arg2, Optional<::ymmsl::SettingValue>());
```
maarten-ic, line 40: Maybe you can simplify this to:
```suggestion
        return BaseMockFun::called_with(arg1, arg2, {});
```
but I don't know if that compiles :)

IrisvdWerf: Unfortunately, this does not compile

maarten-ic: Okay, then keep it as is :) Thanks for checking!

LourensVeen: That's weird, it should. In `libmuscle/cpp/src/libmuscle/util.hpp`:
```
        /** Create a nil optional.
         *
         * This allows you to write {} instead of
         * Optional<LengthyTypeName>() when you need a nil optional.
         */
        Optional(std::initializer_list<T>);
```
What compile error are you getting?

IrisvdWerf: error: no matching function for call to ‘libmuscle::impl::MockGetSetting::called_with(const char*&, const char*&, <brace-enclosed initializer list>) const’
   40 |         return BaseMockFun::called_with(arg1, arg2, {});

maarten-ic: Perhaps it's because the Mock function arguments are a template type, and the compiler doesn't know which type to instantiate unless explicitly provided?

scripts/make_libmuscle_api.py
```diff
@@ -968,6 +968,21 @@ def __copy__(self) -> 'Elements':
...
+        T(), 'get_setting_with_default_as',
+        [String('name'), T('default_value')], True,
+        cpp_chain_call=lambda cpp_args, tpl_type, **kwargs: (
+            f'self_p->get_setting_as<{tpl_type}>({cpp_args})')),
+    OverloadSet('get_setting_with_default', [
+        'get_setting_with_default_as_character',
+        'get_setting_with_default_as_int8',
+        'get_setting_with_default_as_real8',
+        'get_setting_with_default_as_logical'
+        ], False),
```
maarten-ic: Did you leave out the `int8array`, `real8array` and `real8array2` variants on purpose?

libmuscle/cpp/src/libmuscle/instance.hpp
```diff
@@ -793,6 +825,11 @@ class Instance {
 
         Impl const * impl_() const;
         Impl * impl_();
+
+        template <typename ValueType>
+        ValueType get_setting_as_impl_(
+                std::string const & name,
+                Optional<ValueType> const & default_value) const;
```
maarten-ic: This seems unused, could you remove it?

docs/source/fortran.rst
```diff
@@ -268,6 +268,10 @@ Supported types are ``character``, ``logical``, ``int8`` (a 64-bit integer), ``r
 double precision reals). See the :f:func:`API documentation
 <LIBMUSCLE_Instance_get_setting_as_character>` for details.
 
+You can provide a default value as an additional argument, which will be returned
+if the setting is not configured. If no default is provided and the setting is
+missing, an error will be printed and the program will be halted.
```
maarten-ic: In Fortran you need to call a different subroutine (`LIBMUSCLE_Instance_get_setting_with_default`) to supply a default argument.
Can you document the new function as well in the Fortran API docs (`docs/source/fortran_api.rst`)?

maarten-ic (commented):
Looks good! I have a couple of comments left, see below. Let me know if anything's unclear!

libmuscle/cpp/src/libmuscle/instance.hpp
LourensVeen: I'm not sure, but it could be that the Fortran bindings need it. Please check.

IrisvdWerf: Fortran does not depend on get_setting_as_impl  only on get_setting_as. So, I removed it.

docs/source/fortran.rst
```diff
@@ -268,9 +296,11 @@ Supported types are ``character``, ``logical``, ``int8`` (a 64-bit integer), ``r
...
 <LIBMUSCLE_Instance_get_setting_as_character>` for details.
 
-You can provide a default value as an additional argument, which will be returned
-if the setting is not configured. If no default is provided and the setting is
-missing, an error will be printed and the program will be halted.
+If you want to provide a default value to be returned when the setting is not
+configured, you need to call a different function:
+``LIBMUSCLE_Instance_get_setting_with_default_as_<type>()``.  If no default is
+provided and the setting is missing, an error will be printed and the program 
+will be halted.
```
maarten-ic: The `LIBMUSCLE_Instance_get_setting_with_default_as_<type>()` require a default value, right?

```suggestion
``LIBMUSCLE_Instance_get_setting_with_default_as_<type>()``.
```

docs/source/fortran_api.rst
```diff
@@ -1781,6 +1781,98 @@ LIBMUSCLE_Instance
...
+
+    :p LIBMUSCLE_Instance self: The instance to get the setting from.
+    :p character name: The name of the setting to retrieve.
+    :p real default_value: The value to return if the setting does not exist (kind=LIBMUSCLE_real8).
+    :p integer err_code: An error code output (optional).
+    :p character err_msg: An error message output (allocatable, optional).
+    :r value: The setting's value, or the default value (kind=LIBMUSCLE_real8).
+    :rtype value: real
+
+.. f:function:: LIBMUSCLE_Instance_get_setting_with_default_as_logical(self, name, default_value, err_code, err_msg)
```
maarten-ic: Can you also add the vector ones ([int], [float] and [[float]])?

maarten-ic (commented):
Just some small comments on the docs left

maarten-ic (approved):
LGTM!

docs/source/fortran.rst
```diff
@@ -296,6 +296,13 @@ Supported types are ``character``, ``logical``, ``int8`` (a 64-bit integer), ``r
 double precision reals). See the :f:func:`API documentation
 <LIBMUSCLE_Instance_get_setting_as_character>` for details.
 
+If you want to provide a default value to be returned when the setting is not
+configured, you need to call a different function:
+``LIBMUSCLE_Instance_get_setting_with_default_as_<type>()``, which requires a
```
DaanVanVugt: What happens if you provide an interface definition which defines the default value as optional, won't your C++ method will just receive a null pointer there which you can check for?
That may give a slightly nicer interface but could be a little more brittle for older compilers, and may not be worth the hassle.

Some AI slop:
```fortran
    interface
        ! The BIND(C) name must match the C++ function name exactly
        subroutine my_cpp_function(opt_arg) bind(C, name="my_cpp_function")
            import :: c_int
            ! Declare the argument as OPTIONAL
            integer(c_int), intent(in), optional :: opt_arg
        end subroutine my_cpp_function
    end interface
```

DaanVanVugt (approved):

docs/source/fortran.rst
maarten-ic: I agree we could make a nicer Fortran interface, but not without doing work in the bindings generator.

I doubt that your AI code will compile as optional arguments are not compatible with C calling conventions:
C doesn't have default arguments of functions, so in the C interface you would either need two sets of functions (like in this PR) or add a boolean argument that indicates whether a default value was provided.
Then in the Fortran interface you could mark the default value as optional, and choose to call the correct C function based on `if present(default_value)`.

DaanVanVugt: Okay, let's not do it then, it's not that much work to type the extra few characters to select the right function.

docs/source/fortran.rst
```diff
@@ -296,6 +296,13 @@ Supported types are ``character``, ``logical``, ``int8`` (a 64-bit integer), ``r
 double precision reals). See the :f:func:`API documentation
 <LIBMUSCLE_Instance_get_setting_as_character>` for details.
 
+If you want to provide a default value to be returned when the setting is not
+configured, you need to call a different function:
```
LourensVeen: I'm not super happy with this, because I'd really like to keep the API as identical as possible between the supported languages. I think we can fix it with an OverloadSet, but I have to dig into that a bit more since it's been many years since I wrote the API generator script.

I'm happy to pick that up, and I'll just push a commit if I can figure out how to do it.

libmuscle/cpp/src/libmuscle/instance.cpp
```diff
@@ -88,9 +88,15 @@ class Instance::Impl {
         bool reuse_instance();
         void error_shutdown(std::string const & message);
         std::vector<std::string> list_settings() const;
-        ::ymmsl::SettingValue get_setting(std::string const & name) const;
+        ::ymmsl::SettingValue get_setting(
+                std::string const & name,
+                Optional<::ymmsl::SettingValue> const & default_value) const;
         template <typename ValueType>
         ValueType get_setting_as(std::string const & name) const;
```
LourensVeen: For `get_setting()`, the original got replaced with the new two-argument version, but for `get_setting_as()` the new one got added.

I wasn't sure that I wasn't missing something (overload resolution in C++ is a beast) so I've tried removing the one-argument one and setting an empty default for `default_value` in the two-argument version, and that works fine (see commit). So we can save some code here.

libmuscle/cpp/src/libmuscle/instance.hpp
```diff
@@ -269,12 +270,18 @@ class Instance {
...
          * @return The value of the setting as a generic SettingValue.
          *
-         * @throw std::out_of_range if no setting with the given name exists.
+         * @throw std::out_of_range if no setting with the given name exists
+         *                          and no default value was provided.
          */
-        ::ymmsl::SettingValue get_setting(std::string const & name) const;
+        ::ymmsl::SettingValue get_setting(
+                std::string const & name,
+                Optional<::ymmsl::SettingValue> const & default_value = {}) const;
```
LourensVeen: This adds the `Optional` class to the public API, which I've been trying to avoid. (Although I never documented that fact, sorry :sweat:.) The reason for that is that `Optional` is really a temporary thing to tide us over until we can switch to C++17 and use `std::optional`.

We're actually probably at that point already, and I should make the change when I find some time, but that would mean changing the API if `Optional` is in it. So I'd really prefer having an overload here, with a one-argument and a two-argument version side-by-side.

libmuscle/cpp/src/libmuscle/settings_manager.cpp
```diff
@@ -67,7 +67,8 @@ std::vector<std::string> SettingsManager::list_settings(
 
 SettingValue const & SettingsManager::get_setting(
         Reference const & instance,
-        Reference const & setting_name
+        Reference const & setting_name,
+        Optional<SettingValue> const & default_value
```
LourensVeen: I don't really think that we need a `default_value` argument in `SettingsManager`. We could just have one only in `Instance`, from there call `Instance::Impl` and have that calls `SettingsManager`, then catch any `std::out_of_range` exceptions in `Instance` and substitute the `default_value` if needed. That would save some more code.

The only reason not to do that would be performance, because exception handling isn't as efficient in C++ as it is in Python, but if the speed of your simulation depends on the performance of `get_setting()` then something has gone wrong somewhere.

As for the overload, I've tried this out to check that it actually saves code, and it's not a big difference but it does help. So I've pushed that commit too.

LourensVeen (changes requested):
Hi everyone,

I've been looking at this and playing a bit with some ideas, and I think I've identified some further improvements. I actually implemented two, since it's been a while since I've written any C++ and I wasn't sure they would actually work, but they seem to, so I've pushed those commits.

I have to do some work for another project today, but I'll look at the Fortran API to see if we can make it consistent. I agree with Daan that it's a small thing, but I also want to see if it can be done. So I'll pick that up.

I'm not sure if you still have time to work on this today, but if you do then maybe you can look at removing `Optional` from the public API? If not then that's also okay, then I'll pick that up before I merge it.

Unless, of course, you disagree with any of this, in which case please let me know! I may be the maintainer but I'm not omniscient, so I may well have overlooked something. Four always know at least as much as one.

Finally, Iris, nice job! Yes, it can be improved some more I think, but it works and it fits in with the rest and it's nice to read, and that's a very good start. Welcome to the team!

libmuscle/cpp/src/libmuscle/instance.hpp
maarten-ic: That's fine for me. @IrisvdWerf can you pick this up?

docs/source/fortran.rst
maarten-ic: If I'd write this Fortran API by hand, I'd update the `get_setting_as_<type>` to add an optional default parameter and do
```fortran
if present(default) then
    <call C API with default value>
else
    <call C API without default value>
end if
```

The annoying bit is:
- The API already has default error parameters (which most people aren't using). To stay API compatible, however, you'll need to add the default value as the last argument. E.g.
   ```fortran
       subroutine LIBMUSCLE_Instance_get_setting_as_int8array( &
            self, name, value, err_code, err_msg, default)
   ```
  Not a big downside, since users can use keyword arguments in their calls, e.g. `call instance%get_setting_as_int8array('test', settingvalue, default=(/1, 2, 3/))`
- The wrapper generator doesn't have such a pattern.
---
> I think we can fix it with an OverloadSet,

That may work indeed:
- Rename existing `get_setting_as_<type>` Fortran functions to (for example) `get_setting_without_default_as_<type>`
- Keep `get_setting_with_default_as_<type>`
- Add overloadset (Fortran generic) for `get_setting_as_<type>` and let the F compiler choose between the with/without default

At the expense of additional Fortran methods

libmuscle/cpp/src/libmuscle/instance.hpp
maarten-ic: Note that moving to that will revert https://github.com/multiscale/muscle3/pull/340/changes/01d1ef0d4b207af2490c19b69a2f3a4eb6504781 :)

docs/source/fortran.rst
maarten-ic: Do you want to keep API compatibility for the C layer, or is that considered an implementation detail?

LourensVeen: Ah, see, I hadn't considered the error parameters, good point.

The C layer is internal, and ABI compatibility isn't guaranteed, although it seems that our fusion friends do appreciate it because recompiling their codes takes significant effort. But we've never made that promise, and if it needs to be done from time to time then it needs to be done I think.

libmuscle/cpp/src/libmuscle/instance.hpp
LourensVeen: No, that's an internal interface where we can use `Optional`, I think?

maarten-ic: Looking at this in more detail, I think we're both partly right :)
- On the `Instance::Impl` class we can keep the changes from 01d1ef0d4b207af2490c19b69a2f3a4eb6504781
- On the public `Instance` API we need to revert part of the changes to have a one-argument and a two-argument `get_setting_as` method again

LourensVeen: Ah, the hallmark of true collaboration :laughing:. Yes, that makes sense.

LourensVeen: Fixed by @IrisvdWerf, nice!

docs/source/fortran.rst
LourensVeen: I've implemented the overload option, with (internal) function names `get_setting_as1` and `get_setting_as2` (1 and 2 input arguments) and an overload `get_setting_as`. I've added an `OverloadSetTmpl` class to the API generator to generate the overload wrapper. That generator could use some more documentation and some cleanup, but that's for another day :sweat_smile:

I'm a little scared to merge it without someone with more Fortran skills having a look at it, but it passes the tests and so far that's been the standard with respect to Fortran support in MUSCLE3, so here we go :smile:.

LourensVeen (approved):

libmuscle/fortran/src/libmuscle/libmuscle.f90
```diff
@@ -794,20 +808,27 @@ module libmuscle
...
+    public :: LIBMUSCLE_Instance_get_setting_as2_int8array
+    public :: LIBMUSCLE_Instance_get_setting_as2_real8array
+    public :: LIBMUSCLE_Instance_get_setting_as2_real8array2
     public :: LIBMUSCLE_Instance_get_setting_as_character
     public :: LIBMUSCLE_Instance_get_setting_as_int8
     public :: LIBMUSCLE_Instance_get_setting_as_real8
     public :: LIBMUSCLE_Instance_get_setting_as_logical
     public :: LIBMUSCLE_Instance_get_setting_as_int8array
     public :: LIBMUSCLE_Instance_get_setting_as_real8array
     public :: LIBMUSCLE_Instance_get_setting_as_real8array2
```
maarten-ic, line 831: These seem to be missing an `interface` declaration. The member functions are generated correctly though :/

LourensVeen: I think that these declare the generic wrappers public, and those wrappers forward to the actual functions `get_setting_as1` and `get_setting_as2`, which have an `interface` declaration and a C wrapper. But I may have misunderstood something?

maarten-ic: There was still a Fortran interface missing to say that `LIBMUSCLE_Instance_get_setting_as_int8` can call either of `LIBMUSCLE_Instance_get_setting_as1_int8` or `LIBMUSCLE_Instance_get_setting_as2_int8` depending on the arguments.

This was not implemented for the new OverloadSetImpl -> see #342

PR #343, opened 2026-03-13
Add libmuscle version to manager log output

maarten-ic:
This PR adds the libmuscle version to the muscle3_manager.log. This information could be useful for debugging simulation crashes.


LourensVeen (approved):
Good idea, for debugging compatibility issues and also for general provenance.

Issue #314, opened 2025-04-15
Assume `threads: 1` execution for models not explicitly specified in resources section


DaanVanVugt:
We have quite a few of these lines, and it is always `threads: 1`. Perhaps it would be a nice default, instead of raising an error about missing resources section?

maarten-ic:
Continuation of discussion from https://github.com/multiscale/muscle3/pull/345#discussion_r3141709361:

Options to implement:
1. Update ymmsl-python [`Configuration.check_consistent`](https://github.com/multiscale/ymmsl-python/blob/develop/ymmsl/v0_2/configuration.py#L534) to allow empty resource requests for non-MPI programs. Then update muscle3 to default to "1 thread" for any program without resource requests.
2. Update ymmsl-python to create (during `check_consistent()`?) default resources for any non-MPI program that is missing them.

My preference is for option 1, but I'll let @LourensVeen weigh in on this as well.

LourensVeen:
Agreed, 1 is better.

LourensVeen:
This was implemented in #359 and released with 0.10.0, so I'll close it.

PR #345, opened 2026-03-31
Feature/implementation tester

IrisvdWerf:


libmuscle/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -0,0 +1,168 @@
...
+                            f"Expected exactly one component."
+                        )
+                    target_model = model
+                    target_name = comp_name
+                    target_comp = comp
+
+        if target_model is None or target_comp is None or target_name is None:
+            raise ValueError(
+                f"No component found using implementation '{implementation}'."
+            )
```
maarten-ic: Would it be easier to search for any `model` or `program` called `implementation`?

E.g.

```python
# Find the program or submodel that we want to test:
implementation: Implementation  # Model and Program are both subclasses of Implementation
if implementation_name in model.models:
    implementation = model.models[implementation_name]
elif implementation_name in model.programs:
    implementation = model.models[implementation_name]
else:
    raise ValueError(f"No implementation '{implementation}' found in the yMMSL")
```

libmuscle/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -0,0 +1,168 @@
...
+
+        tester_o_i_ports = []
+        tester_s_ports = []
+
+        # F_INIT and S of target (inputs) → tester needs O_I (outputs) to send to them
```
maarten-ic: Could you use [`Ports.sending_port_names()`](https://ymmsl-python.readthedocs.io/en/develop/api.html#ymmsl.v0_2.Ports.sending_port_names) and [`Ports.receiving_port_names()`](https://ymmsl-python.readthedocs.io/en/develop/api.html#ymmsl.v0_2.Ports.receiving_port_names) here?

libmuscle/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -0,0 +1,168 @@
...
+                    conduit = Conduit(
+                        f"{target_name}.{port_name}",
+                        f"tester.receive_{port_name}",
+                    )
+                    target_model.conduits.append(conduit)
+
+        tester_ports = Ports(
+            o_i=tester_o_i_ports if tester_o_i_ports else None,
+            s=tester_s_ports if tester_s_ports else None,
+        )
```
maarten-ic: I'd expect that empty lists are accepted here as well as None:
```suggestion
        tester_ports = Ports(o_i=tester_o_i_ports, s=tester_s_ports)
```

libmuscle/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -0,0 +1,168 @@
...
+
+        tester_comp = Component(
+            name="tester",
```
maarten-ic: Let's call this the same as the implementation?

```suggestion
            name="muscle3_implementation_tester",
```

libmuscle/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -0,0 +1,168 @@
...
+            ports=tester_ports,
+            description="Manual tester component for implementation testing",
+            implementation="muscle3_implementation_tester",
+            optional=False,
+        )
+        target_model.components[Reference("tester")] = tester_comp
```
maarten-ic: Idem here

libmuscle/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -0,0 +1,168 @@
...
+
+        tester_program = Program(
+            name="muscle3_implementation_tester",
+            execution_model=ExecutionModel.MANUAL,
+            description="Manual tester program for implementation testing",
+        )
+        if config.programs is None:
```
maarten-ic: config.Programs shouldn't be None, right?

libmuscle/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -0,0 +1,168 @@
...
+        manager_stdout = self.run_dir / "manager_stdout.txt"
+        manager_stderr = self.run_dir / "manager_stderr.txt"
+
+        with open(manager_stdout, "w") as f_out, open(manager_stderr, "w") as f_err:
+            self._manager_process = subprocess.Popen(
+                ["muscle_manager", "--start-all", str(test_ymmsl_path)],
+                stdout=f_out,
+                stderr=f_err,
+                cwd=str(self.run_dir),
+            )
```
maarten-ic: This makes it more difficult to obtain the connection settings for the muscle_manager.

Could you reuse the logic for starting the muscle manager as a [`multiprocessing.Process`](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Process) from [here](https://github.com/multiscale/muscle3/blob/develop/integration_test/conftest.py#L53-L76)?

libmuscle/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def __enter__(self) -> "MuscleTester":
+        """Allows usage in a with-statement"""
+        return self
+
+    def __exit__(
+        self,
+        typ: type[BaseException] | None,
+        exc: BaseException | None,
+        tb: TracebackType | None,
```
maarten-ic: My bad: this is only supported in newer Python versions (the CI reports an error on it)


```suggestion
        typ: Optional[type[BaseException]],
        exc: Optional[BaseException],
        tb: Optional[TracebackType],
```

libmuscle/python/libmuscle/pytest/implementation_tester.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def __init__(self, default_timeout: float) -> None:
+        """
+        Initialize the implementation tester.
+
+        Args:
+            default_timeout: Default timeout for receive operations in seconds.
+        """
+        self._default_timeout = default_timeout
+        self._instance: Optional[Instance] = None
```
maarten-ic: I'd just create the instance immediately here, then you can remove the `_instance_checked` and `_ensure_initalized` methods

libmuscle/python/libmuscle/pytest/implementation_tester.py
```diff
@@ -0,0 +1,81 @@
...
+    ) -> None:
+        """
+        Send a message on the specified port.
+
+        Args:
+            port_name: Name of the port to send on (without 'send_' prefix).
+            message: The message to send.
+            slot: Optional slot number for vector ports.
+        """
+        actual_port = f"send_{port_name}"
```
maarten-ic: We could make it easier for ourselves by not renaming the ports :)

libmuscle/python/libmuscle/pytest/implementation_tester.py
```diff
@@ -8,34 +10,26 @@ class ImplementationTester:
...
-        This creates the Instance as the 'tester' component which will
-        connect to the manager and discover its ports automatically.
-        """
-        if not self._initialized:
-            self._instance = Instance()
-            self._initialized = True
+        # Pass manager address and instance name through environment
+        os.environ["MUSCLE_MANAGER"] = muscle_manager_address
+        os.environ["MUSCLE_INSTANCE"] = "muscle3_implementation_tester"
+        test_model = test_ymmsl_config.models[
```
maarten-ic: Shouldn't you look in `test_ymmsl_config.programs`?

libmuscle/python/libmuscle/pytest/implementation_tester.py
```diff
@@ -48,12 +42,10 @@ def send(
...
             slot: Optional slot number for vector ports.
         """
-        actual_port = f"send_{port_name}"
-
         if slot is not None:
-            self._instance_checked.send(actual_port, message, slot)
+            self._instance.send(port_name, message, slot)
         else:
-            self._instance_checked.send(actual_port, message)
+            self._instance.send(port_name, message)
```
maarten-ic: https://muscle3.readthedocs.io/en/latest/apidocs/libmuscle.instance.html#libmuscle.instance.Instance.send

If you don't provide `slot`, it will default to None, so the following is equivalent:

```suggestion
        self._instance.send(port_name, message, slot)
```

libmuscle/python/libmuscle/pytest/implementation_tester.py
```diff
@@ -70,12 +62,14 @@ def receive(
             slot: Optional slot number for vector ports.
             timeout: Timeout in seconds. If None, uses default_timeout.
         """
-        actual_port = f"receive_{port_name}"
-
         if timeout is None:
             timeout = self._default_timeout
 
         if slot is not None:
```
maarten-ic: Idem with the `slot` default argument

libmuscle/python/libmuscle/pytest/implementation_tester.py
```diff
@@ -70,12 +62,14 @@ def receive(
             slot: Optional slot number for vector ports.
             timeout: Timeout in seconds. If None, uses default_timeout.
         """
-        actual_port = f"receive_{port_name}"
-
         if timeout is None:
             timeout = self._default_timeout
```
maarten-ic, line 78: Add TODO comment (or indicate in some way that we should still implement something for this)?

libmuscle/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -24,108 +29,90 @@ class MuscleTester:
...
-        exc: BaseException | None,
-        tb: TracebackType | None,
+        typ: Optional[type[BaseException]],
+        exc: Optional[BaseException],
+        tb: Optional[TracebackType],
     ) -> None:
         """Allows usage in a with-statement"""
         self.cleanup()
 
     def add_tester_component(
```
maarten-ic: We don't want users to call this directly, so let's make this a private method?

```suggestion
    def _add_tester_component(
```

libmuscle/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -24,108 +29,90 @@ class MuscleTester:
...
-                        f"{target_name}.{port_name}",
-                        f"tester.receive_{port_name}",
-                    )
-                    target_model.conduits.append(conduit)
-
-        tester_ports = Ports(
-            o_i=tester_o_i_ports if tester_o_i_ports else None,
-            s=tester_s_ports if tester_s_ports else None,
+        tester_model = Model(
+            name=tester_name,
```
maarten-ic: Only name is a mandatory argument, could you simplify to this?

```
tester_model = Model(name=tester_name)
```

libmuscle/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -24,108 +29,90 @@ class MuscleTester:
...
+        for port_name in implementation.ports.sending_port_names():
+            tester_s_ports.append(f"{port_name}")
+            tester_model.conduits.append(
+                Conduit(
+                    f"{implementation_name}.{port_name}",
+                    f"{tester_name}.{port_name}"
+                )
+            )
+
+        tester_model.components[Reference(tester_name)] = Component(
```
maarten-ic, line 109: Shouldn't you add the `implementation_name` to the components as well?

maarten-ic (commented):
Couple of comments, we should probably be close to get this working for a simple case?

libmuscle/python/libmuscle/manager/instance_manager.py
```diff
@@ -127,6 +127,11 @@ def start_all(self) -> None:
                         ', not starting it.')
                 continue
             program = self._configuration.programs[component.implementation]
+            if program.execution_model == ExecutionModel.MANUAL:
+                _logger.info(
+                        f'Instance {instance} has a MANUAL execution model,'
+                        ' not starting it.')
+                continue
```
IrisvdWerf, line 134: Shouldn't this be already implemented in some way?

LourensVeen: I think this is one of the things I've added to ymmsl in support of this project, but haven't implemented yet in MUSCLE3 because I didn't have the time. Conduit filters and timeline annotations on ports are the others. So unless I forgot you're right that it's not yet in there.

IrisvdWerf: No problem, I’ll create a new PR to implement this feature.

integration_test/test_muscle_tester_macro_micro.py
```diff
@@ -0,0 +1,106 @@
...
+def set_pythonpath():
+    """Ensure the codes directory is on PYTHONPATH so 'python3 -m micro' works."""
+    original = os.environ.get('PYTHONPATH', '')
+    codes_path = str(CODES_DIR)
+    if original:
+        os.environ['PYTHONPATH'] = codes_path + ':' + original
+    else:
+        os.environ['PYTHONPATH'] = codes_path
+    yield
+    os.environ['PYTHONPATH'] = original
```
maarten-ic: Pytest has https://docs.pytest.org/en/stable/reference/reference.html#pytest.MonkeyPatch.setenv to do this for you:

```suggestion
def set_pythonpath(monkeypatch):
    """Ensure the codes directory is on PYTHONPATH so 'python3 -m micro' works."""
    monkeypatch.setenv("PYTHONPATH", codes_path, prepend=":")
```

integration_test/test_muscle_tester_macro_micro.py
```diff
@@ -0,0 +1,106 @@
...
+def test_micro_model_with_tester(muscle3_tester: MuscleTester) -> None:
+    """Test the micro model using MuscleTester acting as the macro.
+
+    The micro model:
+      - Receives a value on 'init' (F_INIT)
+      - Sends the same value back on 'final' (O_F)
+
+    The tester (acting as macro):
+      - Sends integers 0 and 1 on 'init'
+      - Expects to receive the same integers back on 'final'
```
maarten-ic, line 28: As long as the test code itself is pretty clear (which it is in this case), I wouldn't put an extensive docstring at the top

maarten-ic: Idem for the other test

integration_test/test_muscle_tester_macro_micro.py
```diff
@@ -0,0 +1,106 @@
...
+CODES_DIR = Path(__file__).parent / 'codes'
+
+
+@pytest.fixture
+def muscle3_tester(tmp_path: Path):
+    """Pytest fixture providing a MuscleTester instance with a run directory."""
+    run_dir = tmp_path / 'run_dir'
+    run_dir.mkdir()
+    with MuscleTester(run_dir) as tester:
+        yield tester
```
maarten-ic: The muscle3_tester fixture should be available (because libmuscle is a pytest plugin) without declaring it here, right?


```suggestion
```

libmuscle/python/libmuscle/pytest/implementation_tester.py
```diff
@@ -61,15 +68,37 @@ def receive(
...
             timeout: Timeout in seconds. If None, uses default_timeout.
+
+        Raises:
+            RuntimeError: If a deadlock is detected while waiting for a message.
         """
         if timeout is None:
             timeout = self._default_timeout
 
-        if slot is not None:
+        prev_timeout = self._instance._communicator._receive_timeout
```
maarten-ic: Do we need to restore the timeout?

libmuscle/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -0,0 +1,187 @@
...
+            self.control_pipe = None
+            self.process = None
+
+
+def start_mmp_server(control_pipe: Tuple[Connection, Connection],
+                     ymmsl_config: Configuration, run_dir: RunDir,
+                     env: Optional[dict] = None) -> None:
+    if env is not None:
+        os.environ.clear()
+        os.environ.update(env)
```
maarten-ic: You don't use the case where `env is None`, so this can be simplified to:
```suggestion
                     env: dict[str, str]) -> None:
    os.environ.clear()
    os.environ.update(env)
```

libmuscle/python/libmuscle/pytest/test/test_muscle_tester.py
maarten-ic, line 1: Let's move this inside a new `pytest/test/` folder? Similar to how all `manager` tests are in the `manager/test` folder.

libmuscle/python/libmuscle/test/test_muscle_tester.py
```diff
@@ -0,0 +1,112 @@
...
+
+def test_add_test_conduits_to_config(tmp_run_dir: Path, program_config: Configuration
+                                     ) -> None:
+    tester = MuscleTester(tmp_run_dir)
+    result = tester._add_tester_component(program_config, "micro_model_program")
+    tester_model = result.models[Reference("muscle3_test_model")]
+    conduit_pairs = {
+        (str(c.sender), str(c.receiver)) for c in tester_model.conduits
+    }
+    # Tester sends init_in → implementation receives init_in
```
maarten-ic: Let's just use ascii characters, instead of unicode arrows
```suggestion
    # Tester sends init_in -> implementation receives init_in
```

Also for the other occurrences in this file.

libmuscle/python/libmuscle/test/test_muscle_tester.py
```diff
@@ -0,0 +1,112 @@
...
+    assert (
+        "muscle3_implementation_tester.init_in",
+        "micro_model_program.init_in",
+    ) in conduit_pairs
```
maarten-ic: Do you need to build a `conduit_pairs` first? It'd be cleaner if you could:
```python
assert Conduit(
    "muscle3_implementation_tester.init_in", "micro_model_program.init_in"
) in tester_model.conduits
```

libmuscle/python/libmuscle/test/test_muscle_tester.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def test_add_test_ports_to_config(tmp_run_dir: Path, program_config: Configuration
+                                  ) -> None:
+    tester = MuscleTester(tmp_run_dir)
+    result = tester._add_tester_component(program_config, "micro_model_program")
+    tester_model = result.models[Reference("muscle3_test_model")]
+    tester_comp = tester_model.components[Reference("muscle3_implementation_tester")]
+    # init_in is a receiving port of micro_model_program → tester sends on it (O_I)
+    assert "init_in" in [str(p) for p in tester_comp.ports.sending_port_names()]
```
maarten-ic:
```suggestion
    assert Identifier("init_in") in tester_comp.ports.sending_port_names()
```

maarten-ic (commented):
I think I've looked at the relevant changes, but I may have missed things (apologies in advance). Can you have a look at the following comments?

libmuscle/python/libmuscle/pytest/muscle_tester.py
IrisvdWerf: I thought this could be an option if the python version < 3.14?

libmuscle/python/libmuscle/pytest/muscle_tester.py
maarten-ic, line 143: Can you add a docstring?

We also need to add a documentation entry in the sphinx docs so people will know that this exists and can be used for testing their implementations.

maarten-ic (commented):
No further comments on the code at this moment -- I'll revisit when #347 is merged.

Let's have a look at documenting this feature?

docs/source/muscle_tester.rst
```diff
@@ -0,0 +1,121 @@
...
+Given a yMMSL file that describes your implementation (its ports, executable,
+etc.), :class:`~libmuscle.pytest.MuscleTester`:
+
+1. Reads the yMMSL configuration and automatically adds a *tester component*
+   that is wired to every port of the implementation under test.
+2. Starts a real MUSCLE3 manager in a background process.
+3. Returns an :class:`~libmuscle.pytest.implementation_tester.ImplementationTester`
+   object that you can use to :meth:`~libmuscle.pytest.implementation_tester.ImplementationTester.send`
+   and :meth:`~libmuscle.pytest.implementation_tester.ImplementationTester.receive`
+   messages on those ports, just as the real coupled component would.
```
maarten-ic: I'm not sure if this belongs in the user documentation.

If you want to include it, probably better to do it all the way at the end. The main goal for this doc page should be: how to use this feature.

docs/source/muscle_tester.rst
```diff
@@ -0,0 +1,121 @@
...
+
+    ymmsl_version: v0.2
+
+    programs:
+      micro:
+        ports:
+          f_init: init
+          o_f: final
+        executable: python3
+        args: "-m micro"
```
maarten-ic: Let's provide an example pointing to a python file instead of a python module?

docs/source/muscle_tester.rst
```diff
@@ -0,0 +1,121 @@
+.. _muscle-tester:
+
+==========================
+Testing your MUSCLE3 model
+==========================
+
+When developing a MUSCLE3 component, it is useful to be able to test it in
+isolation — without having to set up and run the full coupled simulation. The
+:class:`~libmuscle.pytest.MuscleTester` class (together with the
+``muscle3_tester`` pytest fixture) makes this easy: it starts a real MUSCLE3
```
maarten-ic, line 10: Explain pytest and fixture?

docs/source/muscle_tester.rst
```diff
@@ -0,0 +1,121 @@
...
+.. code-block:: python
+    :caption: test_micro.py
+
+    from libmuscle import Message
+    from libmuscle.pytest import MuscleTester
+
+
+    def test_micro_model(muscle3_tester: MuscleTester) -> None:
+        """Test the micro model by acting as the macro."""
+        tester = muscle3_tester.start_implementation("micro.ymmsl", "micro")
```
maarten-ic: This makes some assumptions on the file/directory layout. Can you include that somewhere as well? I'd imagine these files to live in a separate `tests` folder and you'd run `pytest` from the project root folder?

docs/source/muscle_tester.rst
```diff
@@ -0,0 +1,121 @@
...
+    pytest test_micro.py
+
+
+Timeouts and error handling
+============================
+
+By default, :meth:`~libmuscle.pytest.MuscleTester.start_implementation`
+uses a 60-second timeout for all receive operations. If the implementation
+does not send a message within that time, a :class:`RuntimeError` is raised
+and the test fails. You can adjust this timeout:
```
maarten-ic: Explain why you'd want to do this? E.g. long-running calculations?

docs/source/muscle_tester.rst
```diff
@@ -0,0 +1,121 @@
...
+
+You can also override the timeout for individual receive calls:
+
+.. code-block:: python
+
+    reply = tester.receive("final", timeout=2.0)
+
+
+API reference
+=============
```
maarten-ic: API is already generated elsewhere, you can link to it instead of re-generating here.

(probably this will lead to a lot of sphinx warnings about duplicate references?)

maarten-ic: Examples: https://github.com/multiscale/muscle3/blob/develop/docs/source/tutorial.rst?plain=1#L93

docs/source/python_api.rst
```diff
@@ -9,3 +9,11 @@ API Documentation for Python
...
 .. automodule:: libmuscle.runner
    :members: run_simulation
+
+.. automodule:: libmuscle.pytest
+   :members:
+   :undoc-members:
+
+.. automodule:: libmuscle.pytest.implementation_tester
+   :members:
+   :undoc-members:
```
maarten-ic: Do you need both?

docs/source/muscle_tester.rst
maarten-ic: Note that users don't necessarily have `pytest` installed after installing muscle3

maarten-ic (commented):
Couple of comments on the docs

docs/source/muscle_tester.rst
LourensVeen: Is there a use case for inlining the ymmsl? I haven't really thought about how this would be used, but that would solve the need-to-find-that-file problem.

integration_test/ymmsl/codes/macro.ymmsl
```diff
@@ -17,3 +17,7 @@ programs:
       in: Expects a single integer of any value
     executable: python3
     args: "-m macro"
+
+resources:
+  macro:
+    threads: 1
```
LourensVeen: You've mentioned a default 1-thread resources entry in case it's not specified. Yesterday that came up in discussion with the integrated modelling group, and I've decided that we should have one at least for non-MPI models. Testing is a nice use case for it :smile:

libmuscle/python/libmuscle/pytest/implementation_tester.py
```diff
@@ -0,0 +1,101 @@
...
+        os.environ["MUSCLE_MANAGER"] = muscle_manager_address
+        os.environ["MUSCLE_INSTANCE"] = "muscle3_implementation_tester"
+        test_model = test_ymmsl_config.programs[
+            Reference("muscle3_implementation_tester")
+            ]
+        instance_ports = {
+            Operator.O_I: [str(p) for p in test_model.ports.sending_port_names()],
+            Operator.S: [str(p) for p in test_model.ports.receiving_port_names()]
+        }
+        self._instance = Instance(ports=instance_ports)
```
LourensVeen: Should we turn off the MMSF checker here? When it will start checking timelines in the future we're going to have complaints I think.

LourensVeen:
I finally found a bit of time to have a look at what you've been up, and I really like where this is going, it's an elegant idea and a nice clean implementation. Good work!

docs/source/muscle_tester.rst
IrisvdWerf: `start_implementation` currently only accepts a file path string. To support inlining, the method would need to be extended to also accept a `ymmsl.v0_2.Configuration` object directly, right?

docs/source/python_api.rst
IrisvdWerf: No, I think the second one can be removed as the user receives the ImplementationTester instance directly from MuscleTester.start_implementation().

docs/source/muscle_tester.rst
maarten-ic: > start_implementation currently only accepts a file path string. To support inlining, the method would need to be extended to also accept a ymmsl.v0_2.Configuration object directly, right?

It might be nice to support similar arguments as `ymmsl.load_as`:
- If argument is a `str`, it should be a string containing the ymmsl (so the yMMSL can be inline in the test file, see example below)
- If argument is a pathlib.Path, it should point to a file containing the ymmsl

Example:
```python
config = """
ymmsl_version: v0.2
description: yMMSL config for testing xyz implementation
implementations:
  xyz:
    executable: /path/to/xyz
"""

def test_xyz(muscle3_tester):
    tester = muscle3_tester.start_implementation(config, "xyz")
    ...
```

integration_test/ymmsl/codes/macro.ymmsl
IrisvdWerf: I added the resources entry because MuscleTester otherwise crashes with a KeyError in the Planner if it's missing. Introducing a default theads: 1 for non-MPI components would remove the need for this workaround.

I can add this in a new PR. Would we prefer to handle it centrally in the Planner, like a fallback if missing, or earlier during model validation so the configuration is always complete?

maarten-ic: ymmsl-python needs an update for this as well ([`Configuration._check_resources`](https://github.com/multiscale/ymmsl-python/blob/develop/ymmsl/v0_2/configuration.py#L534) should allow empty resources requests for non-MPI execution models)

> Would we prefer to handle it centrally in the Planner, like a fallback if missing, or earlier during model validation so the configuration is always complete?

I think it'd be cleanest to handle it in the Planner (and update `Configuration._check_resources`), but maybe @LourensVeen has another opinion on this?

maarten-ic: Let's continue the discussion in #314

docs/source/muscle_tester.rst
```diff
@@ -0,0 +1,221 @@
...
+    resources:
+      micro:
+        threads: 1
+
+.. note::
+    The path passed to
+    :meth:`~libmuscle.pytest.MuscleTester.start_implementation` is resolved
+    relative to the directory from which you run ``pytest`` — typically the
+    project root. The executable in the yMMSL file is also launched from that
+    directory, so make sure your implementation is importable from there.
```
maarten-ic: Did you test this? I'm not sure if this is correct: in a regular muscle3 simulation the python executable is started inside the run directory.

For these tests that run directory would be somewhere in a temporary folder, and `micro.py` may not be found like this.

docs/source/muscle_tester.rst
```diff
@@ -0,0 +1,221 @@
...
+
+    resources:
+      micro:
+        threads: 1
+    """
+
+Step 2: Use the ``muscle3_tester`` fixture in your test
+-------------------------------------------------------
+
+**Using a Path:**
```
maarten-ic: Might be nice to use tabs for the two cases? See the dynamic port configuration PR for some examples: https://github.com/multiscale/muscle3/pull/350/changes#diff-509659d28e6e781efc52de54ecdf261fd94227761dbc10004b7d3d24a95a4460

docs/source/muscle_tester.rst
```diff
@@ -0,0 +1,221 @@
...
+
+Timeouts and error handling
+============================
+
+By default, :meth:`~libmuscle.pytest.MuscleTester.start_implementation`
+uses a 60-second timeout for all receive operations. If the implementation
+does not send a message within that time, a :class:`RuntimeError` is raised
+and the test fails.
+
+You may want to **increase** the timeout when your implementation performs
```
maarten-ic: No need to emphasize
```suggestion
You may want to increase the timeout when your implementation performs
```

docs/source/muscle_tester.rst
```diff
@@ -0,0 +1,221 @@
...
+expensive work between messages, for example, a micro-model that runs a
+numerical solver for several minutes per time step. In those cases the default
+60 seconds may expire before the component has had a chance to reply, causing a
+spurious test failure.
+
+Conversely, you may want to **decrease** the timeout in fast unit tests so
```
maarten-ic:
```suggestion
Conversely, you may want to decrease the timeout in fast unit tests so
```

docs/source/muscle_tester.rst
```diff
@@ -0,0 +1,221 @@
...
+.. code-block:: python
+
+    reply = tester.receive("final", timeout=2.0)
+
+
+API reference
+=============
+
+The full API for :class:`libmuscle.pytest.MuscleTester` and
+:class:`libmuscle.pytest.implementation_tester.ImplementationTester` is
```
maarten-ic: You could prefix this with `~` so it renders just the final part of the dotted name in the HTML docs.

```suggestion
The full API for :class:`~libmuscle.pytest.MuscleTester` and
:class:`~libmuscle.pytest.implementation_tester.ImplementationTester` is
```

docs/source/muscle_tester.rst
```diff
@@ -0,0 +1,221 @@
...
+:class:`~libmuscle.pytest.MuscleTester` class (together with the
+``muscle3_tester`` pytest fixture) makes this easy: it starts a real MUSCLE3
+manager in the background and connects a *tester component* to all ports of
+your implementation, so you can send and receive messages programmatically from
+within a pytest test.
+
+.. note::
+
+   The testing infrastructure described here relies on `pytest
+   <https://docs.pytest.org>`_, which is **not** installed automatically as
```
maarten-ic:
```suggestion
   <https://docs.pytest.org>`_, which is not installed automatically as
```

docs/source/muscle_tester.rst
IrisvdWerf: You are right, is this solved if you give an absolute path for the script in args?

docs/source/python_api.rst
```diff
@@ -9,3 +9,7 @@ API Documentation for Python
 
 .. automodule:: libmuscle.runner
    :members: run_simulation
+
+.. automodule:: libmuscle.pytest
```
maarten-ic, line 13: This doesn't work correctly for me (when building the docs through `tox -e docs`). I'm getting the following error message in the logs:

```text
WARNING: autodoc: failed to import module 'pytest' from module 'libmuscle'; the following exception was raised:
['Traceback (most recent call last):\n', '  File "/home/maarten/projects/muscle3/muscle3/.tox/docs/lib/python3.12/site-packages/sphinx/ext/autodoc/importer.py", line 269, in import_object\n    module = import_module(modname, try_reload=True)\n             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n', '  File "/home/maarten/projects/muscle3/muscle3/.tox/docs/lib/python3.12/site-packages/sphinx/ext/autodoc/importer.py", line 175, in import_module\n    module = importlib.import_module(modname)\n             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n', '  File "/usr/lib/python3.12/importlib/__init__.py", line 90, in import_module\n    return _bootstrap._gcd_import(name[level:], package, level)\n           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n', '  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import\n', '  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load\n', '  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked\n', '  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked\n', '  File "<frozen importlib._bootstrap_external>", line 995, in exec_module\n', '  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed\n', '  File "/home/maarten/projects/muscle3/muscle3/libmuscle/python/libmuscle/pytest/__init__.py", line 1, in <module>\n    import pytest\n', "ModuleNotFoundError: No module named 'pytest'\n"] [autodoc.import_object]
```

Looks like `pytest` is not installed in the environment for building documentation: can you add it to `tox.ini`?

docs/source/muscle_tester.rst
```diff
@@ -0,0 +1,227 @@
...
+                args: /path/to/my_project/micro.py
+
+            resources:
+              micro:
+                threads: 1
+
+    .. group-tab:: Using an inline string
+
+        You can embed the yMMSL configuration directly in your test file as an
+        f-string. Using ``Path(__file__).parent.resolve()`` gives the absolute path
```
maarten-ic: It doesn't have to be an f-string, any string would do.

```suggestion
        You can embed the yMMSL configuration directly in your test file as a
        string.
        
        Tip: using ``Path(__file__).parent.resolve()`` gives the absolute path
```

libmuscle/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -0,0 +1,221 @@
...
+from libmuscle.manager.manager import Manager
+from libmuscle.manager.run_dir import RunDir
+from libmuscle.receive_timeout_handler import ReceiveTimeoutHandler
+
+
+def raise_error(*args: object) -> None:
+    raise RuntimeError(args)
+
+
+class MuscleTester:
```
maarten-ic, line 35: Could you improve the generated documentation a bit?

1. Perhaps add in the class description that you don't need to construct a `MuscleTester`, but should use the `muscle3_tester` pytest fixture instead?
2. Add a brief docstring to the `cleanup()` function.
3. Fix the layout of `start_implementation`: probably it's sufficient to add a blank line before the first list item.

Image

libmuscle/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -0,0 +1,221 @@
...
+        that a RuntimeError is raised when a receive timeout is reached, causing the
+        test simulation to quit.
+        - The ImplementationTester is then created based on the manager address and the
+        generated test yMMSL configuration.
+
+        Args:
+            ymmsl_source: Either a str containing the yMMSL, or a pathlib.Path
+            pointing to a file containing the yMMSL.
+            implementation: Name of the implementation under test.
+            default_timeout: Timeout (seconds) for message operations (default: 60).
```
maarten-ic:
```suggestion
            ymmsl_source: Either a string containing the yMMSL, or a pathlib.Path
                pointing to a file containing the yMMSL.
            implementation: Name of the implementation to test.
            default_timeout: Timeout (seconds) for message operations.
```

Explanation:
- You need to indent a text block when continuing the description of an argument.
- Default values are documented by looking at the function definition. Putting it in the docstring has a risk that (when updating the function definition) the docstring is not reflecting reality.

maarten-ic (commented):
Couple of suggestions for improving the generated API documentation:

maarten-ic (approved):

docs/source/muscle_tester.rst
```diff
@@ -0,0 +1,219 @@
...
+        string.
+        
+        Tip: using ``Path(__file__).parent.resolve()`` gives the absolute path.
+
+        .. code-block:: python
+            :caption: tests/test_micro.py
+
+            from pathlib import Path
+
+            PROJECT_DIR = Path(__file__).parent.parent.resolve()
```
LourensVeen: or `.parents[1]` to avoid the repetition.

docs/source/muscle_tester.rst
```diff
@@ -0,0 +1,219 @@
...
+Testing your MUSCLE3 model
+==========================
+
+When developing a MUSCLE3 component, it is useful to be able to test it in
+isolation — without having to set up and run the full coupled simulation. The
+:class:`~libmuscle.pytest.MuscleTester` class (together with the
+``muscle3_tester`` pytest fixture) makes this easy: it starts a real MUSCLE3
+manager in the background and connects a *tester component* to all ports of
+your implementation, so you can send and receive messages programmatically from
+within a pytest test.
```
LourensVeen: Maybe point out that this allows automatic testing, as opposed to doing it by hand. To me, that's the main advantage, that you can essentially script the whole testing procedure and easily run it whenever you change something.

docs/source/muscle_tester.rst
```diff
@@ -0,0 +1,219 @@
...
+            CONFIG = f"""
+            ymmsl_version: v0.2
+
+            programs:
+              micro:
+                ports:
+                  f_init: init
+                  o_f: final
+                executable: python3
+                args: {PROJECT_DIR / "micro.py"}
```
LourensVeen, line 146: Should that be TESTS_DIR instead of PROJECT_DIR, or vice versa?

docs/source/muscle_tester.rst
```diff
@@ -0,0 +1,219 @@
...
+Conversely, you may want to decrease the timeout in fast unit tests so
+that a missing ``send`` call is detected quickly rather than making the test
+suite hang for a full minute.
+
+You can adjust this timeout:
+
+.. code-block:: python
+
+    tester = muscle3_tester.start_implementation(
+        TESTS_DIR / "micro.ymmsl", "micro", default_timeout=5.0
```
LourensVeen: Also here TESTS vs PROJECT

docs/source/muscle_tester.rst
```diff
@@ -0,0 +1,219 @@
...
+
+Run ``pytest`` from the project root so that all paths resolve correctly:
+
+.. code-block:: bash
+
+    cd my_project
+    pytest
+
+
+Timeouts and error handling
```
LourensVeen, line 174: I like this, good thinking!

integration_test/ymmsl/codes/macro.ymmsl
LourensVeen: We almost have that feature now. Do you want to merge #359 and then update this?

integration_test/test_muscle_tester_macro_micro.py
```diff
@@ -0,0 +1,85 @@
...
+        )
+
+        reply = Message(msg.timestamp, msg.next_timestamp, msg.data)
+        tester.send('in', reply)
+
+
+def test_receive_timeout_raises_error(muscle3_tester: MuscleTester) -> None:
+    """Test that a RuntimeError is raised when the tester's receive times out.
+
+    The micro model waits to receive on 'final' without sending on 'init', the micro
```
LourensVeen: The tester waits, not the micro model, right?

libmuscle/python/libmuscle/pytest/implementation_tester.py
```diff
@@ -0,0 +1,103 @@
...
+            Operator.O_I: [str(p) for p in test_model.ports.sending_port_names()],
+            Operator.S: [str(p) for p in test_model.ports.receiving_port_names()]
+        }
+        self._instance = Instance(
+                ports=instance_ports,
+                flags=InstanceFlags.SKIP_MMSF_SEQUENCE_CHECKS)
+        # Configure the deadlock-detection timeout to match the default receive
+        # timeout: after `default_timeout` seconds of waiting the manager is
+        # notified, and if a deadlock is detected the simulation is aborted.
+        self._instance._communicator.set_receive_timeout(default_timeout)
```
LourensVeen, line 43: We have a setting `muscle_deadlock_receive_timeout` for this. If possible, it would be cleaner to put that in via the configuration than to access a private member here. It's not a huge deal though, if we change something in there then this test will break and can be fixed. Still, it would be nicer not to break encapsulation.

LourensVeen: Hmm, that wouldn't work on a per-receive basis though...

libmuscle/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -0,0 +1,230 @@
...
+            self._patcher = None
+        if self.control_pipe is not None and self.process is not None:
+            self.control_pipe[0].send(True)
+            self.control_pipe[0].close()
+            self.process.join()
+            self.control_pipe = None
+            self.process = None
+
+
+def start_mmp_server(control_pipe: Tuple[Connection, Connection],
```
LourensVeen, line 195: This and `make_server_process` below are pretty much copy-pasted from integration_test/conftest.py. I like having them here, because they're now part of the functionality we offer rather than merely part of the test harness, but do we need the duplication? Or would deduplicating this perhaps be part of a separate issue on using this feature to implement some or all of the integration tests?

pyproject.toml
LourensVeen, line 28: Is it a good idea to add a dependency on `pytest` here? It's not strictly needed, but we are importing it in some of the files, and I could see e.g. Sphinx or some other tool that tries to load all the files crash due to a missing dependency. I'm not sure actually, this is an actual question not a politely phrased suggestion :smile:

LourensVeen (approved):
This is looking really nice! I have a few comments for you to consider, but I'm going to mark this Approve and let you decide what to do with them (well, the PROJECT_DIR vs. TESTS_DIR thing in the docs needs to be fixed probably).

docs/source/muscle_tester.rst
IrisvdWerf: my_project/
            ├── micro.py
            ├── micro.ymmsl
            └── tests/
                └── test_micro.py

micro.py is located in PROJECT_DIR.

docs/source/muscle_tester.rst
IrisvdWerf: This should be indeed PROJECT

libmuscle/python/libmuscle/pytest/implementation_tester.py
IrisvdWerf: We access _communicator directly because the timeout needs to be set per-receive call (see receive() below), which cannot be done via the muscle_deadlock_receive_timeout configuration setting,  that is read only once at Instance initialisation and applies globally.

libmuscle/python/libmuscle/pytest/muscle_tester.py
IrisvdWerf: I think we could deduplicate this as a follow-up issue.

One possible approach:

* Add `start_instances: bool = True` to `start_mmp_server`
* Move both `start_mmp_server` and `make_server_process` into `libmuscle/pytest/manager_process.py`
* Treat them as plain utility functions shared between `muscle_tester.py` and the integration tests

IrisvdWerf: What do you think @maarten-ic?

maarten-ic: Some thoughts:
- I would do the deduplication in this PR
- Good to add `start_instances` argument, but let's not add the default value (IMO that only makes sense when in the majority of cases you'd use the default, but now you need `True` in 50% and `False` in the other 50% of the cases).
- Fine to move to a separate file but then you should put the cleanup logic inside `make_server_process` (like the implementation in conftest.py), where right now the cleanup is on lines 195-199
- `start_mmp_server` is an implementation detail and `make_server_process` is your utility function that would be used from `conftest.py` and the `MuscleTester`

pyproject.toml
maarten-ic: Well, there's something to be said for either way :) 

In principle we could make it an optional dependency and save everyone that doesn't need the pytest functionality about 15MB of dependency packages.
The same could be said for `matplotlib` though: it is only required for the profiling plots and the examples.

But... optional dependencies are a bit annoying for users when they do want to use the optional functionality. And compared to `numpy` and `matplotlib`, the `pytest` dependency adds just a little extra disk space, so fine to list it as a required dependency.

> I could see e.g. Sphinx or some other tool that tries to load all the files crash due to a missing dependency.

I think most such tools nowadays do lexical analysis instead of blindly importing all python files from a python project, so this shouldn't be a big issue. The API summary generator in Sphinx requires it indeed, but that is handled by the requirements in `tox.ini`

pyproject.toml
```diff
@@ -31,6 +31,7 @@ dependencies = [
     "msgpack>=1,<2",
     "psutil>=5.0.0",
     "parsimonious",
+    "pytest",
```
maarten-ic, line 34: If we add this dependency, then we should remove the note in the documentation page that asks users to `pip install pytest`.

integration_test/conftest.py
```diff
@@ -132,10 +107,8 @@ def run_manager_with_actors(ymmsl_text, tmpdir, actors, expect_success=True):
     cpp_build_dir = libmuscle_dir / 'cpp' / 'build' / 'libmuscle' / 'tests'
     fortran_build_dir = libmuscle_dir / 'fortran' / 'build' / 'libmuscle' / 'tests'
 
-    with ExitStack() as stack:
-        # start muscle_manager and extract manager location
-        ctx = contextmanager(make_server_process)(ymmsl_doc, tmpdir)
-        env['MUSCLE_MANAGER'] = stack.enter_context(ctx)
+    with make_server_process(ymmsl_doc, Path(tmpdir), False) as muscle_manager_address:
+        env['MUSCLE_MANAGER'] = muscle_manager_address
```
maarten-ic: Can you please put `make_server_process` back in the ExitStack?

libmuscle/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -41,8 +42,7 @@ def __init__(self, run_dir: Path) -> None:
         self.run_dir = run_dir
         self.run_dir.mkdir(parents=True, exist_ok=True)
         self.implementation_tester: Optional[ImplementationTester] = None
-        self.control_pipe: Optional[Tuple[Connection, Connection]] = None
-        self.process: Optional[mp.Process] = None
+        self._server_ctx: Optional[Any] = None
```
maarten-ic: Can't you use `AbstractContextManager[str]` instead of `Any` here?

See:
- https://docs.python.org/3/library/typing.html#typing.ContextManager
- https://docs.python.org/3/library/contextlib.html#contextlib.AbstractContextManager

libmuscle/python/libmuscle/pytest/muscle_tester.py
maarten-ic: It might be nicer to use an [ExitStack here](https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack)? It would simplify the cleanup() quite a bit and then we don't need to call the dunder methods `__enter__` and `__exit__` explicitly:

```python
# in __init__:
self._exitstack = ExitStack()
# in start_implementation():
server_ctx = make_server_process(test_ymmsl_config,self.run_dir, True)
muscle_manager_address = self._exitstack.enter(server_ctx)
self._exitstack.enter(patch.object(ReceiveTimeoutHandler, 'on_timeout', raise_error))
self.implementation_tester = ImplementationTester(...)
self._exitstack.callback(self.implementation_tester.cleanup)
# in cleanup()
self._exitstack.close()
self.implementation_tester = None
```

N.B. haven't tested the above code, so please check for correctness :)

libmuscle/python/libmuscle/pytest/muscle_tester.py
maarten-ic: Comment can be ignored when using ExitStack (see the other comment): then we don't need to store the _server_ctx and _patcher anymore (since the exit stack will take care of that).

maarten-ic (commented):
Couple of comments, let me know if you have any questions about these!

libmuscle/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -42,7 +43,7 @@ def __init__(self, run_dir: Path) -> None:
         self.run_dir = run_dir
         self.run_dir.mkdir(parents=True, exist_ok=True)
         self.implementation_tester: Optional[ImplementationTester] = None
-        self._server_ctx: Optional[Any] = None
+        self._exitstack = ExitStack()
         self._patcher: Optional[Any] = None
```
maarten-ic: This member can now be removed right?

maarten-ic (commented):
This should be my final nitpick 😉

maarten-ic (approved):
🚀

PR #347, opened 2026-04-09
Start a program/model manually when ExecutionModel is MANUAL

IrisvdWerf:


libmuscle/python/libmuscle/manager/test/test_instance_manager.py
```diff
@@ -0,0 +1,170 @@
...
+        'libmuscle.manager.instance_manager.NativeInstantiator',
+        MockNativeInstantiator
+    ):
+        instance_manager = InstanceManager(config, run_dir, instance_registry)
+        instance_manager.set_manager_location('localhost:9000')
+        instance_manager.start_all()
+        num_running = instance_manager._num_running
+        instance_manager.shutdown()
+
+    assert num_running == 1
```
IrisvdWerf, line 133: Or do we also want to include counting the running processes which were started manually?

muscle3/muscle_manager.py
```diff
@@ -44,19 +44,20 @@
...
             ' which should be passed to the instances via the'
             ' MUSCLE_MANAGER environment variable or on their command line'
             ' using the --muscle-manager=<contents> option.'
             '\b\n\n'
-            ' If --start-all is not specified and --location-file is also'
-            ' not given, then the location will be printed on standard'
-            ' output.')
+            ' If --start-all is not specified (or if there are manual'
+            ' components) and --location-file is also not given, then the'
+            ' location will be printed on standard output.')
```
maarten-ic: End with another `'\b\n\n'` for consistency?

libmuscle/python/libmuscle/manager/test/test_instance_manager.py
```diff
@@ -0,0 +1,170 @@
...
+from libmuscle.manager.run_dir import RunDir
+from libmuscle.planner.resources import Core, CoreSet, OnNodeResources, Resources
+
+
+def make_configuration(
+        macro_execution_model: ExecutionModel = ExecutionModel.DIRECT,
+        micro_execution_model: ExecutionModel = ExecutionModel.DIRECT
+) -> Configuration:
+    """Create a test configuration with macro and micro components."""
+
```
maarten-ic: Maybe it's cleaner to define the configuration as a yMMSL string and use `load_as(Configuration, ymmsl)` to parse it?

libmuscle/python/libmuscle/manager/test/test_instance_manager.py
```diff
@@ -0,0 +1,170 @@
...
+from libmuscle.manager.instance_manager import InstanceManager
+from libmuscle.manager.instance_registry import InstanceRegistry
+from libmuscle.manager.instantiator import InstantiationRequest
+from libmuscle.manager.run_dir import RunDir
+from libmuscle.planner.resources import Core, CoreSet, OnNodeResources, Resources
+
+
+def make_configuration(
+        macro_execution_model: ExecutionModel = ExecutionModel.DIRECT,
+        micro_execution_model: ExecutionModel = ExecutionModel.DIRECT
```
maarten-ic: Perhaps just leave out these parameters and hardcode that macro has MANUAL and micro has DIRECT execution?

In the tests you're not varying this anyway

libmuscle/python/libmuscle/manager/instance_manager.py
```diff
@@ -127,6 +127,11 @@ def start_all(self) -> None:
                         ', not starting it.')
                 continue
             program = self._configuration.programs[component.implementation]
+            if program.execution_model == ExecutionModel.MANUAL:
```
maarten-ic: This works, but best practice when comparing `Enum`s is to use `is` and `is not` instead of `==` and `!=`, see https://docs.python.org/3/howto/enum.html#comparisons


```suggestion
            if program.execution_model is ExecutionModel.MANUAL:
```

muscle3/muscle_manager.py
```diff
@@ -136,6 +138,11 @@ def _manage_simulation(
 
     run_dir_obj = create_run_dir(run_dir, root_model)
 
+    has_manual_components = any(
+        program.execution_model == ExecutionModel.MANUAL
```
maarten-ic:
```suggestion
        program.execution_model is ExecutionModel.MANUAL
```

docs/source/examples/rd_manual.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+source python/build/venv/bin/activate
+
+rm -f manager_location.txt
+
+DONTPLOT=1 muscle_manager --start-all \
```
maarten-ic, line 7: May be good to add some comments to this file to explain what happens? E.g.
- `# Start muscle_manager in the background, which starts the micro instance`
- `# Wait until the manager has written its network location`
- `# Manually start the macro instance`
- `# Wait for the manager process to complete`

maarten-ic (commented):
Looks good! I have a couple of comments, can you look at them and let me know if you have any questions?

maarten-ic (approved):

libmuscle/python/libmuscle/manager/test/test_instance_manager.py
```diff
@@ -0,0 +1,151 @@
...
+        'libmuscle.manager.instance_manager.NativeInstantiator',
+        MockNativeInstantiator
+    ):
+        instance_manager = InstanceManager(configuration, run_dir, instance_registry)
+        instance_manager.set_manager_location('localhost:9000')
+
+        sent_requests = []
+        original_put = instance_manager._requests_out.put
+
+        def capture_put(item):
```
LourensVeen: Can we just pick the instantiation requests out of `instance_manager._instantiator._requests_queue`? Now we intercept them, then pass them on only for them to end up in the mock we've created above.

libmuscle/python/libmuscle/manager/test/test_instance_manager.py
```diff
@@ -0,0 +1,151 @@
...
+
+class MockNativeInstantiator:
+    """Mock for NativeInstantiator that records requests and returns fake results."""
+
+    def __init__(self, resources_queue, requests_queue, results_queue,
+                 log_records_queue, run_dir):
+        self._resources_queue = resources_queue
+        self._requests_queue = requests_queue
+        self._results_queue = results_queue
+        self._log_records_queue = log_records_queue
```
LourensVeen: Other than `self._resources_queue` these are unused I think? Then they can just be omitted, except for `self._requests_queue` if my comment below is implemented.

LourensVeen (commented):
Very nice! Out MATLAB users will be happy, and probably also people trying to debug their codes by starting them manually from their IDE.

libmuscle/python/libmuscle/manager/test/test_instance_manager.py
IrisvdWerf: Yes, that’s a good point. Rather than intercepting and forwarding via `capture_put`, we now drain `instance_manager._instantiator._requests_queue` directly using the helper `_drain_instantiation_requests`.

libmuscle/python/libmuscle/manager/test/test_instance_manager.py
```diff
@@ -86,37 +85,42 @@ def join(self):
...
 
-def test_start_all_skips_manual_instances(tmp_path, caplog):
+def _drain_instantiation_requests(queue):
+    """
+    Drain all InstantiationRequests currently in the queue.
+    """
+    requests = []
+    while True:
+        try:
+            item = queue.get(timeout=1.0)
```
maarten-ic: Do we need to wait 1 second here?

If the queue is filled in this process/thread (I expect so, but please check), we could also use [`queue.get_nowait()` ](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue.get_nowait).

IrisvdWerf: The test fails when changing `queue.get(timeout=1.0)` to `queue.get_nowait()`. The queue is empty when `_drain_instantiation_requests` is called, the timeout was necessary to wait for the items to arrive.

maarten-ic: Clear, thanks for checking! Maybe nice to reduce the timeout a bit so the test executes faster?

```suggestion
            item = queue.get(timeout=0.1)
```

libmuscle/python/libmuscle/manager/test/test_instance_manager.py
```diff
@@ -0,0 +1,154 @@
...
+        'libmuscle.manager.instance_manager.NativeInstantiator',
+        MockNativeInstantiator
+    ):
+        instance_manager = InstanceManager(configuration, run_dir, instance_registry)
+        instance_manager.set_manager_location('localhost:9000')
+
+        with caplog.at_level(logging.INFO):
+            instance_manager.start_all()
+
+        instantiation_requests = _drain_instantiation_requests(
```
LourensVeen: This seemed more complicated than necessary to me, but I wasn't sure if I was missing something, so I experimented a bit and came up with this, removing `_drain_instantiation_request`:

```suggestion
        request = instance_manager._instantiator._requests_queue.get(timeout=0.1)
        assert isinstance(request, InstantiationRequest)
        assert request.instance == Reference('micro')
        with pytest.raises(Empty):
            instance_manager._instantiator._requests_queue.get(timeout=0.1)
```
Draining only one message is pretty easy :smile:

libmuscle/python/libmuscle/manager/test/test_instance_manager.py
LourensVeen: I thought the same, but the queue is a multiprocessing.queue which has a pipe and a background thread to pump it. Setting the timeout causes `get()` (which in this test is running in the same process and thread as `put()`) to wait, which enables a context switch to the background thread to pump the data.

LourensVeen (commented):
I was hoping that this change would shorten and simplify things, and it looks like that's possible. How about this?

LourensVeen (approved):
Awesome, let's go!

PR #348, opened 2026-04-10
Move from setup.py to pyproject.toml

### LINKED ISSUE #298: Switch to pyproject.toml
It's time. See also #102.

maarten-ic:
This commit changes the following:
- Migrate configuration from setup.py, setup.cfg and .coveragerc to pyproject.toml with some tweaks:
  - Removed mypy overrides that were no longer necessary
- Update VERSION to follow PEP440 (see https://peps.python.org/pep-0440/#developmental-releases)
  - Update Makefiles such that `make` regenerates `version.h` when VERSION is updated
- Use `importlib.metadata.version` to obtain Python package version in runtime
- Replace flake8 with ruff (with E and F lints enabled)
  - Include `muscle3` package in linting checks, and fix three errors
- Remove ymmsl dependency from deps in tox.ini: it's coming in with the pyproject.toml dependencies

Fixes #298, #102


maarten-ic:
Alright, fixing pytest was more tricky than expected...

Anyway, it seems to be working now. Additional changes required:
- Remove empty `libmuscle/python/conftest.py` file, this seemed to confuse pytest a lot (because the `<repo>/libmuscle` folder was shadowing the installed `libmuscle` module...)
- Move to `--import-mode=importlib` to prevent `ImportPathMismatchError`s (this was also needed for #345 (WIP))
- Python 3.14 defaults to the `forkserver` multiprocessing start method, which caused an issue in an integration test (caplog cannot be pickled)

integration_test/test_cpp_mmp_client.py
```diff
@@ -110,9 +110,9 @@ def mock_remove(name: Reference):
 @skip_if_python_only
 def test_mmp_client(log_file_in_tmpdir, tmpdir, caplog):
     # caplog cannot be pickled, causing a crash if we try to pass it to the
-    # process on platforms that spawn. In this case, don't pass it and skip
+    # process on platforms that don't fork. In this case, don't pass it and skip
     # the output check, if it works on one platform it'll work everywhere.
-    pass_caplog = caplog if mp.get_start_method() != 'spawn' else None
+    pass_caplog = caplog if mp.get_start_method() == 'fork' else None
```
LourensVeen: So, as of Python 3.14, are there any platforms we support that _do_ fork? Otherwise this will never be run anymore.

libmuscle/cpp/build/libmuscle/Makefile
```diff
@@ -215,7 +215,7 @@ INSTANTIATE_VERSION_H += s/@PROJECT_VERSION_MINOR@/$(minor_version)/\n
 INSTANTIATE_VERSION_H += s/@PROJECT_VERSION_PATCH@/$(patch_version)/\n
 INSTANTIATE_VERSION_H += s/@PROJECT_VERSION@/$(muscle_version)/\n
 
-$(srcdir)/version.h: version.h.in
+$(srcdir)/version.h: version.h.in $(version_file)
```
LourensVeen, line 218: Nice catch!

pyproject.toml
```diff
@@ -0,0 +1,131 @@
...
+requires = ["setuptools>=64"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "muscle3"
+dynamic = ["version"]
+description = "Version 3 of the MUltiScale Coupling Library and Environment"
+readme = "README.rst"
+authors = [
+  {name = "Lourens Veen", email = "l.veen@esciencecenter.nl"},
```
LourensVeen, line 11: I seem to recall that this only allowed a single entry in the past. It's now called `authors` though, which suggests that multiple entries are possible? If so then we should extend it. I'm not sure what qualifies someone as a coauthor, but I'm pretty sure Maarten should at least be there given his past work on the checkpointing algorithm.

pyproject.toml
```diff
@@ -0,0 +1,131 @@
...
+muscle_manager = "muscle3.muscle_manager:manage_simulation"
+muscle3 = "muscle3.muscle3:muscle3"
+
+[tool.setuptools]
+include-package-data = true
+
+[tool.setuptools.packages.find]
+where = [".", "libmuscle/python"]
+include = ["muscle3", "libmuscle", "libmuscle.*"]
+namespaces = false  # Prevent setuptools from marking /libmuscle as a namespace package
```
LourensVeen, line 57: Wow, I didn't know it could or would do that. Aren't namespace packages deprecated anyway? Sounds like a tricky one to find and fix, nice job!

tox.ini
```diff
@@ -20,9 +19,9 @@ passenv =
     MUSCLE_ENABLE_FORTRAN_MPI
 
 commands =
+    ruff check
     mypy
     pytest {posargs}
-    flake8 libmuscle/python/libmuscle integration_test scripts/
```
LourensVeen, line 25: Same comment as for the ymmsl-python one...

LourensVeen (commented):
Nice job, I'm very happy to see us move into the modern era of Python packaging!

integration_test/test_cpp_mmp_client.py
maarten-ic: Python 3.14 still supports _fork_, but defaults to _forkserver_.

I've updated the test to choose _fork_ when the default is _forkserver_ (it applies only to this test, all other tests still use the platform default). Please have a look and let me know what you think of this.

pyproject.toml
maarten-ic: Multiple entries are possible indeed. I've listed myself as well

pyproject.toml
maarten-ic: Namespace packages have valid use cases, it doesn't look like they're deprecated: https://packaging.python.org/en/latest/guides/packaging-namespace-packages/

For the python packaging it had been easier if I had renamed the root `libmuscle` folder to `src` and move the `muscle3` source folder in the `src/python` folder (as a sibling folder to the libmuscle Python package folder), but I didn't want to do such a big refactoring 😅
Luckily setuptools is pretty configurable and it works now!

tox.ini
maarten-ic: Same answer as for the ymmsl-python one :) 

> My reasoning was: ruff is the fastest of the three, so putting it first will give feedback sooner (and waste less time and resources on the more expensive tests).
> 
> Though I don't mind putting it at the end if that helps your workflow :)

maarten-ic (commented):
Thanks for your review @LourensVeen!

I've updated the fork/forkserver test and responded to your comments.
Note that the logging test asserts in `test_cpp_mpp_client.py` were actually skipped previously, so I've updated the test to fix that as well.

Let me know if there's anything!

LourensVeen (approved):
All good!

integration_test/test_cpp_mmp_client.py
LourensVeen: Yeah, I think this is probably best for now. If "fork" gets removed or becomes problematic then we can always figure out some other way of testing it. Thanks!

pyproject.toml
LourensVeen: We should probably do that at some point, but yeah, one at a time.

LourensVeen:
I checked the package this change generates vs. what we had before, and noticed that we were previously shipping `integration_test` :sweat:. Oops! The idea wasn't really for people to install `muscle3` and then to do `from integration_test.test_all import`...

I'm not sure if it was intentional, but thanks for fixing that!

maarten-ic:
Thanks for merging! Regarding your comment in commit e71a44c95a642129b9baa1eee14208c8f902762d:

> Someone should figure out if MANIFEST.in is still relevant these days.

Yes, `MANIFEST.in` is still relevant, and setuptools looks for it (even by default with pyproject.toml), see  https://setuptools.pypa.io/en/latest/userguide/datafiles.html#include-package-data

PR #349, opened 2026-04-10
Prevent double CI runs in Pull Requests

maarten-ic:
Currently, when a PR is opened and new commits are pushed to the PR, the compatibility and continuous_integration workflows execute twice: once for the `push` trigger, and once for the `pull_request` trigger.

This PR updates the triggers such that the `push` trigger is only activated on the `main` or `develop` branches. Preventing double runs of the workflows.


LourensVeen (approved):
Our planet thanks you for saving some energy!

Issue #239, opened 2023-06-21
Document dynamic ports API


LourensVeen:
We have an API in `libmuscle` for making components with ports defined only in the yMMSL file. It's intended to be used to create generic mappers that can e.g. take any number of inputs and merge them into a list of items.

While the individual functions are documented in the API docs, there's no explanation of what they're for and how to use them. We now have users that want to use this functionality, so we should add some documentation on how to do it.

maarten-ic:
Documentation was added in #350: https://muscle3.readthedocs.io/en/develop/dynamic_port_configuration.html

Let's close this issue?

PR #350, opened 2026-04-15
Allow dynamic O_I and S ports

### LINKED ISSUE #344: Allow dynamic O_I and S ports
Currently MUSCLE3 allows for dynamic ports (though undocumented, see #239), which will dynamically create F_INIT/O_F ports for incoming respectively outgoing conduits.
With yMMSL v0.2 it is mandatory to list the ports and their operators in the yMMSL configuration. This means that MUSCLE3 could also start supporting dynamic O_I and S ports.

maarten-ic:
This PR allows components to have dynamic O_I and S ports (in addition to dynamic F_INIT and O_F ports) based on the yMMSL configuration.

Fixes #344, #239

Main changes:
- muscle_manager sends ports configured in the yMMSL file to each instance
- Instances (Python / C++) use the port configuration instead of the conduit list for determining which ports to create
  - This allows specifying dynamic O_I and S ports in addition to F_INIT and O_F.
- Add documentation and example code for dynamic components


libmuscle/python/libmuscle/peer_info.py
```diff
@@ -29,36 +30,44 @@ def __init__(self, kernel: Reference, index: List[int],
...
 
-        self.__peer_dims = peer_dims    # indexed by kernel id
-        self.__peer_locations = peer_locations  # indexed by instance id
+        self._peer_dims = peer_dims    # indexed by kernel id
+        self._peer_locations = peer_locations  # indexed by instance id
+        self._ymmsl_ports = ymmsl_ports
+
+    def list_ymmsl_ports(self) -> list[Port]:
+        """List ports declared in the yMMSL configuration"""
+        return self._ymmsl_ports
```
IrisvdWerf, line 70: Do you need this function, as it is just returning self._ymmsl_ports?

libmuscle/python/libmuscle/test/test_peer_info.py
```diff
@@ -20,7 +20,7 @@ def peer_info() -> PeerInfo:
     peer_locations = {
             Ref('other'): ['tcp:other']}
 
-    ymmsl_ports = [Port(Id("out"), Operator.O_F), Port(Id("in"), Operator.F_INIT)]
+    ymmsl_ports = [Port(Id("out"), Operator.O_I), Port(Id("in"), Operator.S)]
```
IrisvdWerf, line 23: Why would you change the ports of peer_info and peer_info2?

docs/source/examples/cpp/combiner.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include <ymmsl/ymmsl.hpp>
+
+using libmuscle::Data;
+using libmuscle::Instance;
+using libmuscle::Message;
+using ymmsl::Operator;
+
+
+int main(int argc, char * argv[]) {
+    // 1. Request dynamic port configuration by not providing a ports description
```
IrisvdWerf, line 17: What is  meant by ports description?

libmuscle/python/libmuscle/peer_info.py
maarten-ic: Not strictly needed, but I chose to do it like this to stay in line with the other logic in this class.
For example, `get_peer_dims(peer_kernel)` also just returns `self._peer_dims[peer_kernel]`.

libmuscle/python/libmuscle/test/test_peer_info.py
maarten-ic: I found it more logical to see the `kernel` component as the "macro" and the `other` as "micro" due to the ordering of the Conduits in the configuration.

In the end it doesn't really matter much.

docs/source/examples/cpp/combiner.cpp
maarten-ic: The ports description is the declaration of ports when creating an instance. This class is named `PortsDescription` in the C++ interface, that's why I named it like this.

Do you have a suggestion for a better name?

maarten-ic (commented):
.

docs/source/examples/cpp/combiner.cpp
IrisvdWerf: I see, no I think PortsDescription is fine!

IrisvdWerf (approved):

libmuscle/python/libmuscle/test/test_instance.py
```diff
@@ -19,7 +20,7 @@ def logger():
 def MMPClient():
     with patch('libmuscle.instance.MMPClient') as MMPClient:
         mmp_client = MMPClient.return_value
-        mmp_client.request_peers.return_value = (MagicMock(), MagicMock(), MagicMock())
+        mmp_client.request_peers.return_value = PeerInfo(*[MagicMock()]*6)
```
LourensVeen, line 23: Hahaha, that does look magical indeed :smile: With the brackets and asterisks for pixie dust. And still valid Python too! 🪄

libmuscle/cpp/src/ymmsl/component.cpp
```diff
@@ -33,6 +33,21 @@ std::string operator_name(Operator op) {
     throw std::logic_error("Unreachable code reached");
 }
 
+Operator operator_for_name(std::string name) {
```
LourensVeen: Ah finally, a nit! I was starting to feel like I was shirking my duties here. Can we pass this by const reference, that being idiomatic?

libmuscle/cpp/src/libmuscle/tests/test_instance.cpp
```diff
@@ -334,14 +334,8 @@ TEST_F(libmuscle_instance_base, create_instance_profiling) {
...
 TEST_F(libmuscle_instance_base, create_instance_connecting) {
-    auto & rpret = MockMMPClient::return_value.request_peers.return_value.get();
-    auto & peer_dims = std::get<1>(rpret);
-    peer_dims["component"] = {};
-    peer_dims["other"] = {};
-
-    auto & peer_locs = std::get<2>(rpret);
-    peer_locs["component"] = {"tcp:localhost:9003"};
-    peer_locs["other"] = {"tcp:localhost:9004"};
+    auto const mock_peer_info = \
```
LourensVeen, line 337: This return-a-PeerInfo just cleans things up everywhere, including here. Very nice!

docs/source/dynamic_port_configuration.rst
```diff
@@ -0,0 +1,63 @@
...
+
+        .. code-block:: C++
+
+            Instance instance(argc, argv, {
+                {Operator::F_INIT, {"initial_state"}},  // 1D Grid
+                {Operator::O_F, {"final_state"}}});     // 1D Grid
+
+
+Some components in your simulation can be generic, for example merging multiple inputs
+into a single container. For these generic component, MUSCLE3 allows the component to
```
LourensVeen: components

docs/source/examples/cpp/combiner.cpp
```diff
@@ -0,0 +1,66 @@
...
+    Instance instance(argc, argv);
+    // Optionally provide instance flags with: Instance(argc, argv, instance_flags)
+
+    while (instance.reuse_instance()) {
+        // 2. Request which ports are available:
+        auto ports = instance.list_ports();
```
LourensVeen: This should go before the reuse loop I think, not inside it. The ports aren't going to change while we run. So we can get the ports, do the checks, and create `ports` and `f_init_ports` for use within the reuse loop.

docs/source/examples/cpp/combiner.cpp
```diff
@@ -0,0 +1,66 @@
...
+        std::sort(f_init_ports.begin(), f_init_ports.end());
+        for (auto const & port : f_init_ports) {
+            if (instance.is_connected(port)) {
+                input_messages.push_back(instance.receive(port));
+            }
+        }
+
+        // Check that we have at least one input
+        if (input_messages.empty()) {
+            std::string msg = "The combiner actor requires at least one connected F_INIT port.";
```
LourensVeen: This could also be checked before the reuse loop.

docs/source/examples/cpp/combiner.cpp
```diff
@@ -0,0 +1,66 @@
...
+            instance.error_shutdown(msg);
+            throw std::runtime_error(msg);
+        }
+
+        // 4. Combine the input and send on connected O_F ports
+        auto timestamp = input_messages[0].timestamp();
+        auto next_timestamp = input_messages[0].next_timestamp();
+        Data data = Data::nils(input_messages.size());
+        for (std::size_t i = 0; i < input_messages.size(); ++i) {
+            Data copy;
```
LourensVeen, line 59: I'm a bit confused here:

1) Why do we want a copy? We're not going to modify that input data, so we can just take a reference to it for use when sending and rely on the automatic memory management to keep it around
2) Why are we creating a Data containing a nil and then pointing it to the input data? Might as well use the copy constructor then and save a few lines?

docs/source/examples/python/combiner.py
```diff
@@ -0,0 +1,44 @@
...
+from libmuscle import Instance, InstanceFlags, Message
+
+
+def main():
+    # 1. Request dynamic port configuration by not providing a ports description
+    instance = Instance()
+    # Optionally provide instance flags with: Instance(flags=...)
+
+    while instance.reuse_instance():
+        # 2. Request which ports are available:
```
LourensVeen: Same comment as for the C++ version.

docs/source/examples/python/combiner_test.py
```diff
@@ -0,0 +1,40 @@
...
+    print("Starting destination component")
+    instance = Instance({Operator.F_INIT: ["input"]})
+
+    while instance.reuse_instance():
+        msg = instance.receive("input")
+        # Test expected messages that we received
+        assert msg.timestamp == 0.0
+        assert isinstance(msg.data, list)
+        for i in range(len(msg.data)):
+            # source1 will send 1, sourc2 will send 2, etc.
```
LourensVeen: sourc -> source

LourensVeen (changes requested):
This is a very nice addition, and the implementation also looks nice. New feature, some nice refactoring, what more could you want?

I do have some comments about the example that I'd like to see addressed, could you have a look at those?

docs/source/examples/cpp/combiner.cpp
maarten-ic: Good question! The main issue was that `Message.data()` is a `DataConstRef` and we're not allowed to assign those to a `Data` object.

By creating a list of nils, and then making a copy of the data like this, I managed to get it to work. But it does feel quite hacky indeed... Let's discuss if you want?

maarten-ic:
CI is failing due to #360

docs/source/examples/cpp/combiner.cpp
LourensVeen: Okay, discussed, it's still ugly and arguably a loophole, because I think this still allows modifying what should be a constant object. But it's the only way to make it work with the current Data/DataConstRef, and the real solution is probably to track constness inside of the Data object along with the rest of the type, and that's surely a separate issue. So let's leave this for what it is and try to pick up that fix later.

LourensVeen (approved):
All good, just needs another merge from develop to get the #360 fix and the CI to pass.

PR #353, opened 2026-05-01
Relax tox version constraint in .readthedocs.yaml

maarten-ic:
`tox` v3.x needs doesn't work with pyproject.toml files by default, which caused RTD builds to fail.


LourensVeen (approved):
Nice, thanks!

PR #359, opened 2026-05-07
Add a default thread for Non-MPI simulations

IrisvdWerf:
If a non-MPI component does not have resources specified, a default of 1 thread will be assigned. 

It is allowed to not have to specify your resources in a non-MPI component: [#PR 40](https://github.com/multiscale/ymmsl-python/pull/40).

### DIFF
--- a/libmuscle/python/libmuscle/manager/instance_manager.py
+++ b/libmuscle/python/libmuscle/manager/instance_manager.py
@@ -139,7 +139,7 @@ class InstanceManager:
             stdout_path = idir / 'stdout.txt'
             stderr_path = idir / 'stderr.txt'
 
-            res_req = self._configuration.get_resources(component.name)
+            res_req = self._configuration.get_resources(model.name + component.name)
 
             request = InstantiationRequest(
                     instance, program,
--- a/libmuscle/python/libmuscle/manager/instance_manager.py
+++ b/libmuscle/python/libmuscle/manager/instance_manager.py
@@ -6,7 +6,7 @@ from typing import Dict, List, Optional, Tuple, Union
 from multiprocessing import Queue
 from queue import Empty
 
-from ymmsl.v0_2 import Configuration, ExecutionModel, Reference
+from ymmsl.v0_2 import Configuration, ExecutionModel, Reference, ThreadedResReq
 
 from libmuscle.errors import ConfigurationError
 from libmuscle.manager.instance_registry import InstanceRegistry
@@ -139,9 +139,14 @@ class InstanceManager:
             stdout_path = idir / 'stdout.txt'
             stderr_path = idir / 'stderr.txt'
 
+            res_req = self._configuration.resources.get(component.name)
+            if res_req is None:
+                # No resources defined; use default of 1 thread for DIRECT components.
+                res_req = ThreadedResReq(component.name, 1)
+
             request = InstantiationRequest(
                     instance, program,
-                    self._configuration.resources[component.name],
+                    res_req,
                     resources, idir, workdir, stdout_path, stderr_path)
             _logger.info(f'Instantiating {instance}')
             self._requests_out.put(request)
--- a/libmuscle/python/libmuscle/manager/instance_manager.py
+++ b/libmuscle/python/libmuscle/manager/instance_manager.py
@@ -6,7 +6,7 @@ from typing import Dict, List, Optional, Tuple, Union
 from multiprocessing import Queue
 from queue import Empty
 
-from ymmsl.v0_2 import Configuration, ExecutionModel, Reference, ThreadedResReq
+from ymmsl.v0_2 import Configuration, ExecutionModel, Reference
 
 from libmuscle.errors import ConfigurationError
 from libmuscle.manager.instance_registry import InstanceRegistry
@@ -18,6 +18,7 @@ from libmuscle.manager.run_dir import RunDir
 from libmuscle.native_instantiator.native_instantiator import NativeInstantiator
 from libmuscle.planner.planner import Planner, ResourceAssignment
 from libmuscle.planner.resources import Resources
+from libmuscle.util import get_resources
 
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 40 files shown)


libmuscle/python/libmuscle/mcp/tcp_util.py
```diff
@@ -53,7 +53,7 @@ def recv_all(socket: SocketType, length: int) -> bytes:
 
         received_count += received_now
 
-    return databuf
+    return bytes(databuf)
```
IrisvdWerf: This is a fix for: 
libmuscle/python/libmuscle/mcp/tcp_util.py:56: error: Incompatible return value type (got "bytearray", expected "bytes")  [return-value]

I also get this on the develop branch.

maarten-ic: Looks like this is because mypy was updated, see: https://mypy.readthedocs.io/en/stable/changelog.html#enable-strict-bytes-by-default

Could you put this change in a separate PR? --> [#PR 360](https://github.com/multiscale/muscle3/pull/360)

I'd propose to change the returntype from `bytes` to `bytearray` instead: `bytes(databuf)` will create an unnecessary copy of the data.

libmuscle/python/libmuscle/planner/planner.py
```diff
@@ -516,13 +516,27 @@ def allocate_all(
...
+        # Build requirements: filling in defaults for DIRECT components
+        # that have no resources defined (default: 1 thread).
+        requirements: Dict[Reference, ResourceRequirements] = dict(
+                configuration.resources)
+        for component in model.components():
+            if (component.name not in requirements and
+                    component.implementation is not None and
+                    component.implementation in programs and
+                    programs[component.implementation].execution_model
+                    is ExecutionModel.DIRECT):
```
maarten-ic: This is checked already by `Configuration.check_consistent`, right?

libmuscle/python/libmuscle/manager/instance_manager.py
```diff
@@ -139,9 +139,14 @@ def start_all(self) -> None:
             stdout_path = idir / 'stdout.txt'
             stderr_path = idir / 'stderr.txt'
 
+            res_req = self._configuration.resources.get(component.name)
+            if res_req is None:
+                # No resources defined; use default of 1 thread for DIRECT components.
+                res_req = ThreadedResReq(component.name, 1)
```
maarten-ic: Ideally this default is only implemented once (in case this needs changing in the future).

What do you think of:
1. Add a new method in `util.py`, `get_resources(configuration: Configuration, name: Reference) -> ResourceRequirements`. You can include the debug log statement for the default value here as well.
2. Use this method both here and in planner.py.

N.B. In planner.py you could do
```python
requirements: dict[Reference, ResourceRequirements] = {
    component_name: get_resources(configuration, component_name)
    for component_name in model.components
}
```

libmuscle/python/libmuscle/planner/planner.py
IrisvdWerf: `Configuration.check_consistent()` guarantees that any `component.implementation` exists in either `programs` or `models`, so it can be simplified.

libmuscle/python/libmuscle/manager/instance_manager.py
LourensVeen: Hmm, should that `get_resources()` method be in ymmsl-python?

IrisvdWerf: I added the function to libmuscle.util, as it is only adding this default number of threads to be able to run the simulation. However, you add something to a yMMSL object. So, you could also add this to ymmsl-python.

maarten-ic: > Hmm, should that `get_resources()` method be in ymmsl-python?

That may be more logical indeed, especially since [ymmsl-python already mentions that](https://github.com/multiscale/ymmsl-python/pull/40/changes#diff-565e63b3075880569adba3309c030af3106373f6fe0682bee0fd486ccf858311R538):

> For non-MPI components, resources are optional: if not specified, a default of 1 thread will be assumed at runtime (e.g. by muscle3).

maarten-ic (commented):
Looks good! But let's move `get_resources()` to the `ymmsl.v0_2.Configuration` class?

maarten-ic:
Thanks! You'll need to update the dependencies in `pyproject.toml` to point to the develop branch of ymmsl-python (after merging https://github.com/multiscale/ymmsl-python/pull/40 ) so CI will be green again.

IrisvdWerf:
CI failing due to #360

pyproject.toml
```diff
@@ -32,8 +32,8 @@ dependencies = [
     "psutil>=5.0.0",
     "parsimonious",
     "numpy>=1.22",
-    "ymmsl>=0.16.0,<0.17",         # Also in examples requirements.txt
-    # "ymmsl @ git+https://github.com/multiscale/ymmsl-python.git@develop",
+    # "ymmsl>=0.16.0,<0.17",         # Also in examples requirements.txt
```
maarten-ic, line 36: Don't forget to update the `docs/source/examples/requirements.txt` as well 😉

libmuscle/python/libmuscle/planner/planner.py
```diff
@@ -517,13 +517,20 @@ def allocate_all(
...
         root_model = configuration.root_model()
         model = ModelGraph(root_model)
-        requirements = configuration.resources
         programs = configuration.programs
         exclusive = {
                 i for c in model.components() for i in c.instances()
                 if (c.implementation and
                     not programs[c.implementation].can_share_resources)}
 
+        # Build requirements: filling in defaults for DIRECT components
```
LourensVeen: This comment describes what the called function does. That should be (and is) in the called function's docstring. If we decide to change the logic and update `Configuration.get_resources()`. will we remember to check every caller and update comments? If we don't then we'll have a misleading comment and that's worse than no comment...

LourensVeen (approved):
Nice, thanks!

maarten-ic (approved):

PR #360, opened 2026-05-08
mypy no longer treats bytearray values as assignable to the bytes type

IrisvdWerf:
mypy was updated, see: 
https://mypy.readthedocs.io/en/stable/changelog.html#enable-strict-bytes-by-default

"mypy no longer treats bytearray and memoryview values as assignable to the bytes type."

### DIFF
--- a/libmuscle/python/libmuscle/communicator.py
+++ b/libmuscle/python/libmuscle/communicator.py
@@ -189,7 +189,7 @@ class Communicator:
         profile_event.stop()
         if port.is_vector():
             profile_event.port_length = port.get_length()
-        profile_event.message_size = len(encoded_message)
+        profile_event.message_size = len(memoryview(encoded_message))
         if not isinstance(message.data, ClosePort):
             self._profiler.record_event(profile_event)
 
@@ -268,7 +268,7 @@ class Communicator:
         recv_decode_event = ProfileEvent(
                 ProfileEventType.RECEIVE_DECODE, ProfileTimestamp(), None,
                 port, None, slot, port.get_num_messages(),
-                len(mpp_message_bytes))
+                len(memoryview(mpp_message_bytes)))
         mpp_message = MPPMessage.from_bytes(mpp_message_bytes)
         recv_decode_event.stop()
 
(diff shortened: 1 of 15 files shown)


LourensVeen:
I ran into this as well, and looked at changing the type like this, and then there were a lot of changes. I think that what we want is a more generic type, but it seems that that's a bit of a mess. [ByteString](https://docs.python.org/3/library/typing.html#typing.ByteString) is close but deprecated, and I think maybe [Buffer](https://docs.python.org/3/library/collections.abc.html#collections.abc.Buffer) is the thing, but that's only available from Python 3.12...

maarten-ic:
> I ran into this as well, and looked at changing the type like this, and then there were a lot of changes. I think that what we want is a more generic type, but it seems that that's a bit of a mess. [ByteString](https://docs.python.org/3/library/typing.html#typing.ByteString) is close but deprecated, and I think maybe [Buffer](https://docs.python.org/3/library/collections.abc.html#collections.abc.Buffer) is the thing, but that's only available from Python 3.12...

The changes in this PR are rather limited (and all in internal API), so I'm not so concerned about it.
If you want a more generic type, then we could use `Union[bytes, bytearray]` and put it as a type alias perhaps?

maarten-ic (approved):

LourensVeen:
We're actually taking these `bytearray` objects and giving them to msgpack, which according to https://github.com/msgpack/msgpack-python/issues/448 actually takes a `typing.Buffer`. So that's probably the right type to use. It's available from Python 3.12, but `typing_extensions` has it, so we can get it from there and be covered on older but still supported Pythons as well.

I'm going to push that change and then merge this, because this silly thing is holding up everything else.

LourensVeen:
Looking at this some more, it seems that `collections.abc.Buffer` will be the right type to use when Python 3.11 drops out of support.

Besides the `bytearray`s that we're receiving, I've also updated the `bytes` objects that we're sending to `Buffer` for consistency. Note that we're now creating `memoryview` objects in a few places to get the length of the buffer. As the name implies, these don't copy the data, essentially `memoryview` just accesses the (native) buffer object so that we can get at its properties, in this case its length.

PR #364, opened 2026-05-28
Muscle tester respect the default timeout and improved error statements

IrisvdWerf:


libmuscle/python/libmuscle/pytest/implementation_tester.py
```diff
@@ -101,3 +106,25 @@ def cleanup(self) -> None:
         if not self._is_shut_down:
             while self._instance.reuse_instance():
                 pass
+        else:
+            # error_shutdown may have failed if the peer was already gone, so the
+            # MPP server threads and/or the profiler thread may still be running.
```
maarten-ic: Can't you handle this at the point that the error occurs?

I also wouldn't touch the internals of the instance here: if another component gets added in the future that needs to be shut down, then this logic will be outdated. Could you fix this on the instance instead?

libmuscle/python/libmuscle/pytest/implementation_tester.py
```diff
@@ -101,3 +106,25 @@ def cleanup(self) -> None:
...
+            try:
+                mpp_server = self._instance._communicator._server
+                post_office = mpp_server._post_office
+                with post_office._outbox_lock:
+                    for outbox in post_office._outboxes.values():
+                        outbox.deposit(b'')
+                for transport_server in mpp_server._servers:
+                    transport_server.close(graceful=False)
+            except Exception:
+                pass
```
maarten-ic: I don't like a blanket `except Exception`, this is very likely to mask issues with the logic

integration_test/test_muscle_tester_macro_micro.py
```diff
@@ -84,3 +86,39 @@ def test_receive_timeout_raises_error(muscle3_tester: MuscleTester) -> None:
...
+    start = time.monotonic()
+    with pytest.raises(RuntimeError):
+        tester.receive('o_f')
+    elapsed = time.monotonic() - start
+
+    print(
+        f"Expected reconnection retries to last at least {_RECONNECT_TIMEOUT} s,"
+        f" and took {elapsed:.1f} s"
+    )
+    assert elapsed >= max(_RECONNECT_TIMEOUT, default_timeout), (
```
maarten-ic: This seems weird to me? I would expect a reverse check instead: that an error is raised without having to wait for minutes.

N.B. This test now takes more than a minute, which leads to timeouts in the CI.

libmuscle/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -25,13 +25,37 @@
...
+from libmuscle.mcp.tcp_transport_client import TcpTransportClient, _RECONNECT_TIMEOUT
 from libmuscle.receive_timeout_handler import ReceiveTimeoutHandler
+from libmuscle.util import Retrier
 
 
 def raise_error(*args: object) -> None:
     raise RuntimeError(args)
 
 
+def make_connection_error_handler(default_timeout: float) -> Callable[..., None]:
```
maarten-ic: Wouldn't it be easier to patch the _RECONNECT_TIMEOUT value itself?

libmuscle/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -25,13 +25,37 @@
...
+    """Create a replacement for TcpTransportClient._handle_disconnect.
+
+    A lost TCP connection triggers reconnection retries for up to _RECONNECT_TIMEOUT.
+    During testing we want reconnection retries to be bounded by at least 
+    _RECONNECT_TIMEOUT seconds, or default_timeout if that is longer.  The returned
+    handler uses a Retrier initialised with max(_RECONNECT_TIMEOUT, default_timeout) so
+    that it gives up (and raises ConnectionError) once that deadline has been reached.
+    """
+    retrier = Retrier(max(_RECONNECT_TIMEOUT, default_timeout))
```
maarten-ic: Don't you want the smallest of the two here?

libmuscle/python/libmuscle/pytest/implementation_tester.py
```diff
@@ -86,9 +87,12 @@ def receive(
...
             _logger.error(
                 "ImplementationTester: error while waiting for a message on"
-                " port '%s': %s. Shutting down.", port_name, exc
+                " port '%s'. Shutting down.", port_name
             )
-            self._instance.error_shutdown(str(exc))
+            try:
+                self._instance.error_shutdown(str(exc))
+            except RuntimeError:
+                self._instance.error_shutdown(str(exc), graceful=False)
```
maarten-ic: This doesn't seem like a good public API for `error_shutdown`, let's discuss.

libmuscle/python/libmuscle/pytest/__init__.py
```diff
@@ -8,8 +8,29 @@
 __all__ = ["ImplementationTester", "MuscleTester"]
 
 
+def _print_error_logs(run_dir: Path) -> None:
+    """Print ERROR-level lines from the manager log file and show its path.
```
maarten-ic: I'd like to see this more prominent in the pytest output: let's discuss.

Can we also:
1. Mention (at the start) that this is content from muscle3_manager.log
2. Indicate that the log files for the instances that are tested can be found in the "run_dir/instances/<instance_name>" folder

libmuscle/python/libmuscle/pytest/__init__.py
```diff
@@ -8,8 +8,29 @@
...
+
+
 @pytest.fixture
 def muscle3_tester(tmp_path: Path) -> Iterator[MuscleTester]:
     """Pytest fixture providing a MuscleTester instance."""
-    with MuscleTester(tmp_path / "run_dir") as tester:
+    run_dir = tmp_path / "run_dir"
+    with MuscleTester(run_dir) as tester:
         yield tester
+    _print_error_logs(run_dir)
```
maarten-ic: Do we need this output when the test failed?

integration_test/test_muscle_tester_macro_micro.py
```diff
@@ -84,3 +86,37 @@ def test_receive_timeout_raises_error(muscle3_tester: MuscleTester) -> None:
...
+    """
+    default_timeout=1.0
+    tester = muscle3_tester.start_implementation(
+        """
+        ymmsl_version: v0.2
+        programs:
+          test_program:
+            ports:
+              o_f: output
+            executable: python
```
maarten-ic, line 106: TODO: Also test invalid executable (seems to take a long time)

Last lines of pytest failure report (when changing `executable: python` to `executable: pythonX`), which took ~10 minutes before a failure was reported:
```text
        if response[0] == ResponseType.PENDING.value:
>           raise RuntimeError('Timeout waiting for peers to appear')
E           RuntimeError: Timeout waiting for peers to appear

.tox/py312/lib/python3.12/site-packages/libmuscle/mmp_client.py:273: RuntimeError
```

libmuscle/python/libmuscle/pytest/__init__.py
maarten-ic: https://stackoverflow.com/a/71275959

libmuscle/python/libmuscle/pytest/__init__.py
```diff
@@ -8,29 +8,38 @@
...
+    """Collect ERROR- and CRITICAL-level lines from the manager log and instance log
+    locations.
 
     Args:
-        run_dir: The run directory containing muscle3_manager.log.
+        run_dir: The run directory containing muscle3_manager.log and instances.
     """
     log_file = run_dir / "muscle3_manager.log"
 
+    lines = [f"Contents from: {log_file}", ""]
```
maarten-ic: Indicate that this is just the ERROR/CRITICAL lines from the file, not the full contents?

libmuscle/python/libmuscle/pytest/__init__.py
```diff
@@ -8,29 +8,38 @@
...
-        if " ERROR " in line
+        if " ERROR " in line or " CRITICAL " in line
     ]
-
     if error_lines:
-        for line in error_lines:
-            print(f"\n[muscle3_tester] {line}")
-    print(f"\n[muscle3_tester] Full log: {log_file}")
+        lines.extend(error_lines)
+    lines += [
```
maarten-ic: This is more of a style preference, but now you're using both `list.extend` and `+=`. Could be cleaned up:

```python
    error_lines = [
        line for line in log_file.read_text().splitlines()
        if " ERROR " in line or " CRITICAL " in line
    ]
    lines = [
        f"Contents from: {log_file}",
        "",
        *error_lines,
        "",
        "Instance log files can be found in:",
        f"  {run_dir / 'instances'}",
    ]
```

libmuscle/python/libmuscle/pytest/__init__.py
```diff
@@ -8,29 +8,38 @@
...
+                   ) -> Iterator[MuscleTester]:
     """Pytest fixture providing a MuscleTester instance."""
     run_dir = tmp_path / "run_dir"
+
     with MuscleTester(run_dir) as tester:
         yield tester
-    _print_error_logs(run_dir)
+
+    report = _collect_log_report(run_dir)
+    request.node.add_report_section("teardown", "muscle3 tester log", report)
```
maarten-ic, line 45: Nice! 😄

libmuscle/python/libmuscle/communicator.py
```diff
@@ -427,7 +427,10 @@ def _drain_incoming_port(self, port_name: str) -> None:
         port = self._port_manager.get_port(port_name)
         while port.is_open():
             # TODO: log warning if not a ClosePort
-            self.receive_message(port_name)
+            try:
+                self.receive_message(port_name)
+            except RuntimeError:
+                break
```
maarten-ic: Please add when this error may be raised and why we're ignoring it.
Might be nice to add a log message instead of silently suppressing the exception?

libmuscle/python/libmuscle/instance.py
```diff
@@ -185,7 +185,13 @@ def __init__(
         """Stores pre-received messages for f_init ports"""
 
         self._register()
-        self._connect()
+        try:
+            self._connect()
+        except Exception:
+            # Initialisation failed after threads were already started (e.g.
+            # Profiler._communicate and TcpTransportServer.serve_forever).
```
maarten-ic:
```suggestion
            # Clean up when we cannot connect to our peers. This could happen when peers
            # are not started (in time) in a pytest context, but we still want to run
            # more test cases without threads lingering around.
```

libmuscle/python/libmuscle/communicator.py
maarten-ic: I think this also applies to `_drain_incoming_vector_port`, right? So maybe it is better to move this to `_close_incoming_ports`?

libmuscle/python/libmuscle/communicator.py
```diff
@@ -461,10 +458,18 @@ def _close_incoming_ports(self) -> None:
...
+                            self._drain_incoming_port(port_name)
+                        else:
+                            self._drain_incoming_vector_port(port_name)
+                    except RuntimeError:
+                        _logger.warning(
+                                "Could not drain incoming port '%s': connection with"
+                                " peer was lost. This can happen when a peer crashed"
+                                " after start_implementation (e.g. in a pytest"
+                                " context)." )
+                        break
```
maarten-ic: I think we can try to drain the remaining ports still, couldn't we?

```suggestion
```

maarten-ic (approved):

maarten-ic:
Hmm, looks like CI is failing now, but not related to the latest changes: can you have a look at that? Perhaps you'll need to increase some timeouts to prevent this from happening again?

maarten-ic (approved):

integration_test/test_muscle_tester_macro_micro.py
```diff
@@ -79,8 +82,76 @@ def test_receive_timeout_raises_error(muscle3_tester: MuscleTester) -> None:
...
         tester.receive('final')
+
+def test_failing_actor(muscle3_tester: MuscleTester) -> None:
+    """Test that a RuntimeError is raised when the actor crashes after registering.
+
+    The test_program registers with MUSCLE3 using start_implementation, then crashes.
+    Any subsequent communication attempt should raise a RuntimeError. This RuntimeError
+    should be raised in at least min(_RECONNECT_TIMEOUT, default_timeout) seconds. 
+    """
+    default_timeout=1.0
```
LourensVeen: Hmm, why is the linter not complaining about the lack of spaces here? Separate issue I guess, but LourensLint 43.2 flagged it.

libmuscle/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -25,6 +25,8 @@
 from libmuscle.pytest.implementation_tester import ImplementationTester
 from libmuscle.manager.manager import Manager
 from libmuscle.manager.run_dir import RunDir
+from libmuscle.mcp.tcp_transport_client import _RECONNECT_TIMEOUT
```
LourensVeen: This should become public then I think, i.e. `RECONNECT_TIMEOUT` instead of with an underscore. I'm okay with accessing private stuff from a test, but this isn't a test.

libmuscle/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -169,13 +176,21 @@ def start_implementation(
         ymmsl.save(test_ymmsl_config, test_ymmsl_path)
 
         server_ctx = make_server_process(
-            test_ymmsl_config,self.run_dir, True)
+            test_ymmsl_config, self.run_dir, True)
```
LourensVeen, line 179: LourensLint 43.2 approves!

libmuscle/python/libmuscle/communicator.py
```diff
@@ -458,10 +458,17 @@ def _close_incoming_ports(self) -> None:
...
-                    else:
-                        self._drain_incoming_vector_port(port_name)
+                    try:
+                        if not port.is_vector():
+                            self._drain_incoming_port(port_name)
+                        else:
+                            self._drain_incoming_vector_port(port_name)
+                    except RuntimeError:
+                        _logger.warning(
+                                "Could not drain incoming port '%s': connection with"
```
LourensVeen: Currently, users know nothing about draining ports, and therefore this could be confusing. Maybe we should just say `Connection with peer '%sa was lost at the end of the simulation, probably because it crashed.`?

LourensVeen (changes requested):
Nice! I think this is pretty much done, and most of my comments are trivial, but I would like to have the improved error message before we merge this and make a new release.

integration_test/test_muscle_tester_macro_micro.py
IrisvdWerf: The linter isn't complaining because `integration_test/` is not included in ruff's scope. In `pyproject.toml`, the include setting is:
`include = ["pyproject.toml", "muscle3/**/*.py", "libmuscle/python/libmuscle/**/*.py"]`.

IrisvdWerf: Do we want to add `integration_test` to this include?

LourensVeen: I'd say yes, but let's make that a separate issue/PR.

libmuscle/python/libmuscle/communicator.py
```diff
@@ -465,10 +465,8 @@ def _close_incoming_ports(self) -> None:
                             self._drain_incoming_vector_port(port_name)
                     except RuntimeError:
                         _logger.warning(
-                                "Could not drain incoming port '%s': connection with"
-                                " peer was lost. This can happen when a peer crashed"
-                                " after start_implementation (e.g. in a pytest"
-                                " context).", port_name)
+                                "Connection with peer '%sa' was lost at the end of the "
```
LourensVeen: Sorry, I got this wrong, the port name isn't the name of the peer, so now this message will look funny. Can we easily get the name of the peer here? If not, then we should probably say "Connection on port ... was lost..."

Oh, and I think `%sa` should be `%s`?

integration_test/test_muscle_tester_macro_micro.py
IrisvdWerf: I will make a separate PR for this.

LourensVeen (changes requested):
Sorry, one more improvement?

LourensVeen (commented):
Great, nice work! Let's get this out to the users!

LourensVeen (approved):

PR #365, opened 2026-06-05
Update github actions

maarten-ic:
Updates the following actions to a newer version:
- actions/checkout: v4 -> v6
- actions/setup-python: v4 -> v6
- actions/cache: v3 -> v5

Change is required because the old versions used Node.js 20, which will be unsupported after September.


LourensVeen (approved):
Hard to argue with, thanks!

PR #366, opened 2026-06-05
Add all python directories for ruff checks in pyproject.toml

IrisvdWerf:


pyproject.toml
```diff
@@ -114,7 +114,14 @@ exclude = [
   "docs",
   ".git",
 ]
-include = ["pyproject.toml", "muscle3/**/*.py", "libmuscle/python/libmuscle/**/*.py"]
+include = [
```
maarten-ic: What if we'd just remove the include and work with ruff's [default include rules](https://docs.astral.sh/ruff/configuration/#python-file-discovery)?

> By default, Ruff will discover files matching `*.py`, `*.pyi`, `*.ipynb`, or `pyproject.toml`. In [preview](https://docs.astral.sh/ruff/preview/) mode, Ruff will also discover `*.pyw` by default.

IrisvdWerf: That works!

In [#PR364,](https://github.com/multiscale/muscle3/pull/364) we did not get the suggestion to change `default_timeout=0.1` to `default_timeout = 0.1`. With the current settings this is still not been suggested, as ruff does not look at the blank spaces? We could use the [preview setting](https://docs.astral.sh/ruff/preview/) to add this. What do you think?

maarten-ic: We probably need to use ruffs formatter to get that kind of feedback: `ruff format --check`. But then we'll need to apply all of the black/ruff formatting rules, which is better to do after the next release.

maarten-ic (approved):

pyproject.toml
LourensVeen: Agreed, I'll have a go at that once I get it out the door. Let's see how `ruff format` stacks up against LourensLint...

LourensVeen (approved):
Great, thanks!

PR #368, opened 2026-06-14
Update release process post 0.10.0

LourensVeen:
This updates the release documentation with the things I ran into when releasing 0.10.0.

One issue we still have is that I have to temporarily disable branch protection to merge the release branch back into develop, and another is that we can't use setuptools-scm because develop doesn't have the version tags in its past. I'm going to think about whether we can tweak the release process to fix that, and maybe to automate it a bit more, but that's for another time.

### DIFF
--- a/docs/source/releasing.rst
+++ b/docs/source/releasing.rst
@@ -85,8 +85,8 @@ shows up:
 It may give some warnings about missing references, that's a known issue and
 normally harmless. Next, point your web browser to
 ``docs/build/index.html`` and verify that the documentation built
-correctly. In particular, the new version number should be in the browser's
-title bar as well as in the blue box on the top left of the page.
+correctly. Also check that the new version number is in the browser's
+title bar.
 
 Run tests
 ---------
--- a/docs/source/releasing.rst
+++ b/docs/source/releasing.rst
@@ -84,7 +84,7 @@ shows up:
 
 It may give some warnings about missing references, that's a known issue and
 normally harmless. Next, point your web browser to
-``docs/build/html/index.html`` and verify that the documentation built
+``docs/build/index.html`` and verify that the documentation built
 correctly. In particular, the new version number should be in the browser's
 title bar as well as in the blue box on the top left of the page.
 
--- a/docs/source/releasing.rst
+++ b/docs/source/releasing.rst
@@ -67,7 +67,7 @@ Update version
 
 Next, the version should be updated. There is a single version tag in the
 ``VERSION`` file in the root of the repository. On the development branch, the
-version should be set to ``x.y.z-dev``, where ``x.y.z`` is the next expected
+version should be set to ``x.y.z-dev0``, where ``x.y.z`` is the next expected
 version (it's fine if that changes later, e.g. because you end up releasing
 2.0.0 rather than 1.4.0).  On the release branch, it should be set to the number
 of this release of course.
@@ -201,7 +201,7 @@ that we want to have back on the develop branch. So we'll merge it back in:
 
 We use --no-commit to give ourselves a chance to edit the changes before
 committing them. Make sure that README.rst is taken from the develop side,
-and that VERSION is given a new number, probably x.y.{z+1}-dev unless you have big
+and that VERSION is given a new number, probably x.y.{z+1}-dev0 unless you have big
 plans. When done, commit the merge and continue developing.
 
 Update issues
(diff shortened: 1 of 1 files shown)


docs/source/releasing.rst
```diff
@@ -67,7 +67,7 @@ Update version
 
 Next, the version should be updated. There is a single version tag in the
 ``VERSION`` file in the root of the repository. On the development branch, the
-version should be set to ``x.y.z-dev``, where ``x.y.z`` is the next expected
+version should be set to ``x.y.z-dev0``, where ``x.y.z`` is the next expected
```
maarten-ic:
```suggestion
version should be set to ``x.y.z.dev0``, where ``x.y.z`` is the next expected
```

which is the normalized PEP-440 notation (https://peps.python.org/pep-0440/#development-release-separators)

docs/source/releasing.rst
```diff
@@ -206,9 +201,8 @@ that we want to have back on the develop branch. So we'll merge it back in:
 
 We use --no-commit to give ourselves a chance to edit the changes before
 committing them. Make sure that README.rst is taken from the develop side,
-CHANGELOG.md comes from the release branch, and VERSION is given a new number,
-probably x.y.{z+1}-dev unless you have big plans. When done, commit the merge
-and continue developing.
+and that VERSION is given a new number, probably x.y.{z+1}-dev0 unless you have big
```
maarten-ic:
```suggestion
and that VERSION is given a new number, probably x.y.{z+1}.dev0 unless you have big
```

maarten-ic (commented):
> One issue we still have is that I have to temporarily disable branch protection to merge the release branch back into develop

Right, the branch protection rules require that there's an approved Pull Request before allowing merges.

> and another is that we can't use setuptools-scm because develop doesn't have the version tags in its past.

I'm not sure if `setuptools-scm` will be very easy for muscle3, since the VERSION also needs to be included in the C++ (and Fortran) libraries.

> I'm going to think about whether we can tweak the release process to fix that, and maybe to automate it a bit more, but that's for another time.

I'm happy to discuss this with you some time

docs/source/releasing.rst
LourensVeen: Ah, drat. I actually vaguely remember there being some kind of issue with PEP440 and semantic versioning disagreeing on what was and wasn't allowed. But PEP440 compliance is probably more important, as that's what the tooling uses.

docs/source/releasing.rst
LourensVeen: Fixed that. GitHub should really have attached your name to that commit (which I made by clicking the button here) but oh well...

LourensVeen:
About the VERSION file: in AMUSE I'm using hatchling with versioningit instead of setuptools and setuptools-scm, and versioningit lets you write plugins. I made one of those that gets the version from git if we're in a repo, and from a central VERSION file if we're in a tarball. The tarballs you download from GitHub are made with `git archive`, and that has a hook that lets you substitute the version into a file. Pretty neat altogether.

We could maybe do something similar here, but we'd need two implementations, one for the Python side maybe as a setuptools-scm plugin if it supports that, and one in the Makefile that does the same but puts the result into the C++ header. Well, not the highest-priority issue I guess :smile:.

maarten-ic:
> About the VERSION file: in AMUSE I'm using hatchling with versioningit instead of setuptools and setuptools-scm, and versioningit lets you write plugins. I made one of those that gets the version from git if we're in a repo, and from a central VERSION file if we're in a tarball. The tarballs you download from GitHub are made with `git archive`, and that has a hook that lets you substitute the version into a file. Pretty neat altogether.
> 
> We could maybe do something similar here, but we'd need two implementations, one for the Python side maybe as a setuptools-scm plugin if it supports that, and one in the Makefile that does the same but puts the result into the C++ header. Well, not the highest-priority issue I guess 😄.

setuptools-scm has support for [.git_archival.txt](https://setuptools-scm.readthedocs.io/en/latest/usage/#builtin-mechanisms-for-obtaining-version-numbers), and you could run `python -m setuptools_scm` from the Makefile and put that as version in a C++ header as well. So definitely doable!

But then we'd still get the same issue with github and diverging main and develop branches, so the tags are not in the history of the develop branch. That could be fixed by changing the release process slightly to:
1. Create release branch
2. Merge release branch to main
3. Add tag to main and create github release, python packages, etc.
4. Merge main back to develop

Either way, not the highest priority now :)

maarten-ic (approved):

LourensVeen:
Yeah, I was thinking along those lines as well. I'll make a separate issue for that, and I'm going to merge this so we can continue. Thanks!

PR #369, opened 2026-06-14
Add more ruff formatting

LourensVeen:
This enables the `UP` and `B` issue categories in ruff, after which it accuses us of a great many crimes. Being the most guilty developer of all, I went and fixed most of them, and successfully argued with the judge in the remaining ones.

An overview of the types of problems encountered:

### UP
- `Dict`, `List`, `Tuple`, `Set`, `Type` deprecated in favour of their lower-case counterparts
- Various imports from typing that should be from collections.abc
- Use of `.format()` on string literals where an f-string could be used
- Extraneous parentheses in generator expressions
- Extraneous `'r'` mode in `open()` calls
- `functools.cache` is more efficient than` lru_cache` and equivalent if size is infinite
- `%`-formatting is deprecated, but I left those in as they match the `logging` docs exactly

### B
- Unused loop variables
- `@cache` potential memory leak (pleaded innocence, no actual harm done)
- Better warnings by setting stacklevel
- Function call in argument default (`InstanceFlags(0)`)

I also tried to enable `SIM`, but disagreed with most of what it said, so I've left it off. I tried `I` (isort) as well, and like it well enough, but it uncovered a circular import, and I wanted to fix that by moving a class out of `snapshot.py` and then realised that I don't like the use of inheritance in there. So I'm leaving that off for now as well, maybe I'll do some refactoring when I get the time. (Note that I'm not sure why the circular import doesn't show up currently, but we'll take it for now :smile:)

### DIFF
--- a/integration_test/cluster_test/conftest.py
+++ b/integration_test/cluster_test/conftest.py
@@ -339,7 +339,7 @@ def hwthread_to_core():
     a function that takes a comma-separated list of hwthread ids and returns a list of
     corresponding core ids.
     """
-    with open('/proc/cpuinfo', 'r') as f:
+    with open('/proc/cpuinfo') as f:
         cpuinfo = f.readlines()
 
     def get_values(cpuinfo, field):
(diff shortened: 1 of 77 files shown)


maarten-ic (approved):
The world is a much safer place right now 😉 Thanks!

I had a look at the import error, and managed to work around it with just a single `# isort:skip`: I'll create a PR for that one and let you judge if that one is okay!

PR #370, opened 2026-06-15
Enable `isort` rules for linting

maarten-ic:
This PR enables the "I" (isort) rules in ruff.

N.B. there's a single `isort:skip` rule in `__init__.py` to avoid a circular import.


LourensVeen (approved):
We'll see how well my editor folds all those vertical lists, but I guess it makes it easier to figure out which one is triggering the linter if needed. And anyway, it's standard!

PR #377, opened 2026-06-19
Apply ruff format

maarten-ic:
This PR:
1. Applies `ruff format` to the repository
2. Adds a `.git-blame-ignore-revs` to allow to ignore that commit when doing a git blame
3. Adds a step to the tox config to run `ruff format --check` to keep the code formatted according to ruff's rules


LourensVeen (approved):
Yeah, time to say goodbye to LourensLint and its non-standardness and inconsistencies. I'm going to have to learn to type double quotes :smile:.

PR #385, opened 2026-07-27
Stop test when any process fails

IrisvdWerf:


integration_test/conftest.py
```diff
@@ -168,19 +169,37 @@ def run_manager_with_actors(ymmsl_text, tmpdir, actors, expect_success=True):
...
+                    returncode = proc.exitcode
+                if returncode is None:
+                    still_running.append(proc)
+                elif returncode != 0:
+                    failed_processes.append(proc)
+            remaining_processes = still_running
+            if failed_processes:
+                break
+            if remaining_processes:
+                time.sleep(0.1)
```
maarten-ic: Could be slightly simpler when you:
- Make `remaining_processes` a `set` and remove finished processes from the set: this way you don't need a still_running temporary variable
- Make `failed_processes` a boolean (and perhaps rename to has_failed_processes), you're not using the contents of the list
- Remove `if failed_processes: break` and put the condition in the `while remaining_processes and not has_failed_processes`
- Put `time.sleep(0.1)` at the top of the while loop

maarten-ic (approved):
Looks good! Some suggestions to make the code a bit simpler

maarten-ic (approved):
Thx!

integration_test/conftest.py
```diff
@@ -168,19 +169,32 @@ def run_manager_with_actors(ymmsl_text, tmpdir, actors, expect_success=True):
...
+
+        for proc in remaining_processes:
+            proc.terminate()
+        for proc in remaining_processes:
+            if isinstance(proc, subprocess.Popen):
+                proc.wait()
+            else:
+                proc.join()
+
+        if expect_success:
```
LourensVeen: Could be a one-liner?
```assert not has_failed_processes if expect_success else has_failed_processes```
or even
```assert expect_success != has_failed_processes```
but that's probably a bit too obscure :smile:.

I suppose we could replace `has_failed_processes` with `success`, starting at `True` and setting it to `False` if a process crashes. Then we could write the test as
```assert expect_success == success```
and that may just work?

LourensVeen (approved):
Looks good, whichever option you think is best for that final check is fine with me.

PR #386, opened 2026-07-28
Feature/msgpack

SimonPinches:
This patch prevents `msgpack` being downloaded if it's supplied.  (Note, `msgpack` does not come with a `.pc` file).  Also switched to use `c++17` standard to fix `googletest` >= 1.16.

### DIFF
--- a/libmuscle/cpp/build/Makefile
+++ b/libmuscle/cpp/build/Makefile
@@ -31,6 +31,7 @@ dep_name := msgpack
 dep_version_constraint := >= 7.0.0
 dep_version := 7.0.0
 dep_pkgconfig_name := msgpack-cxx
+dep_header := include/msgpack.hpp
 dep_install := 1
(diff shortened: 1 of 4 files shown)


SimonPinches:
I've applied these changes as a patch here: https://github.com/easybuilders/easybuild-easyconfigs/pull/26276

LourensVeen:
Note: I originally understood this to be about avoiding vendoring, which doesn't make sense since MessagePack is now header-only, but the actual rationale is that on HPC, software is often compiled on a worker node, and on some machines those don't have access to the internet, resulting in a build failure.

I was hoping to replace our `Optional` class template with `std::optional` as part of any move to C++17, but that will have to wait. C++17 support is very widespread now and other projects are moving or have already moved, so I think we can do that.

I'm working on some other things, but I'll resolve the conflicts and get this merged before the next release. Thanks!

PR #389, opened 2026-08-07
Don't remove `version.py` with `make clean`

maarten-ic:
Also cleans up sphinx `conf.py`: `version.py` always exists nowadays, since it's tracked by git. This change also applies a ruff format and fixes some (but not all) ruff complaints in `conf.py`.


LourensVeen (approved):
Nice, thanks!

PR #390, opened 2026-08-07
Migrate repository to a `src` layout

### LINKED ISSUE #388: pytest coverage reports 0% coverage for everything but the test modules
### Reproduction

run `tox` (or check a recent CI run) and note that all non-test modules have 0% coverage:

Image

### Cause

This is likely an artifact of using `--import-mode importlib` with pytest, see for example https://github.com/pytest-dev/pytest-cov/issues/677

### Solution direction

We need to use `--import-mode importlib` due to the non-standard code layout (`libmuscle/python/libmuscle` is the root of the python package). We can solve these issues by moving to a `src` layout.

- [ ] Rename `/libmuscle` folder to `/src`
- [ ] Move `/muscle3` into the `/src/python` subfolder
- [ ] Update `Makefile` for new folder structure
- [ ] Update `tox.ini` and `pyproject.toml` configurations for pytest, mypy, ruff, coverage, etc. to use the new folder structure
- [ ] Review documentation and update any references accordingly

---

This would be a relatively straightforward refactoring, but best performed when we don't have open PRs to avoid difficult merge conflicts.

maarten-ic:
Fixes #388

Some notes:
1. `make test_examples` fails (but apparently is not run in the CI) -> created #391 for investigation next week.
2. Moving to a `src` layout was not sufficient to fix #388. libmuscle is a plugin of pytest, so it is imported before the `pytest-coverage` plugin starts recording coverage (skewing the coverage measurements). Applied the workaround suggested by the pytest-cov documentation: https://pytest-cov.readthedocs.io/en/stable/plugins.html.
3. Apparantly `ruff` was a bit confused previously about which modules contained first-party imports (it's `libmuscle.*` and `muscle3.*`). This is now corrected, but lead to some import statements swapping places.


LourensVeen (approved):
Looks nice, and thanks for figuring out all the pytest/coverage stuff, good to have that work again :smile:

PR #393, opened 2026-08-11
Fix `make test_examples` and make it run on pull requests

### LINKED ISSUE #391: Examples are not run in CI
`make test_examples` is not executed in CI.

It's also failing after merging #378 -> to be investigated next week.

maarten-ic:
Fixes made for the examples:

- Handle resizable ports in the TimelineManager
  - Also ensured that the C++ (Sub)TimelineManager was looking at references of the ports stored by the PortsManager instead of a copy.
- Update "load balancer" examples to conform to the Timeline restrictions:    
    1. F_INIT is first (this is covered by the pre_receive of Instance)
    2. All slots of the O_I port (back_out) must send once
    3. All slots of the S port (back_in) must receive once
    4. Repeat 2&3 until done
    5. Send all results on O_F

This PR also enables the `ci_ubuntu24.04` workflow for every pull request, and pushes to the `develop` and `main` branches

Fixes #391

### DIFF
--- a/docs/source/examples/cpp/load_balancer.cpp
+++ b/docs/source/examples/cpp/load_balancer.cpp
@@ -1,5 +1,6 @@
 #include <cstdlib>
 #include <cinttypes>
+#include <limits>
 #include <random>
 
 #include <libmuscle/libmuscle.hpp>
@@ -45,7 +46,8 @@ void load_balancer(int argc, char * argv[]) {
             for (int i = 0; i < num_workers; ++i) {
                 auto msg = (started < num_calls) ? 
                     instance.receive_with_settings("front_in", started)
-                    : Message(-INFINITY);
+                    // Send a message indicating nothing left to do:
+                    : Message(-std::numeric_limits<double>::infinity(), Settings({{"skip", true}}));
                 instance.send("back_out", msg, i);
                 ++started;
             }
--- a/docs/source/examples/cpp/load_balancer.cpp
+++ b/docs/source/examples/cpp/load_balancer.cpp
@@ -34,22 +34,28 @@ void load_balancer(int argc, char * argv[]) {
     while (instance.reuse_instance()) {
         // F_INIT
         int started = 0;
-        int done = 0;
+        std::vector<Message> result;
 
         int num_calls = instance.get_port_length("front_in");
         int num_workers = instance.get_port_length("back_out");
 
         instance.set_port_length("front_out", num_calls);
 
-        while (done < num_calls) {
-            while ((started - done < num_workers) && (started < num_calls)) {
-                auto msg = instance.receive_with_settings("front_in", started);
-                instance.send("back_out", msg, started % num_workers);
+        while (result.size() < num_calls) {
+            for (int i = 0; i < num_workers; ++i) {
+                auto msg = (started < num_calls) ? 
+                    instance.receive_with_settings("front_in", started)
+                    : Message(-INFINITY);
+                instance.send("back_out", msg, i);
                 ++started;
             }
-            auto msg = instance.receive_with_settings("back_in", done % num_workers);
-            instance.send("front_out", msg, done);
-            ++done;
+            for (int i = 0; i < num_workers; ++i) {
+                result.push_back(instance.receive_with_settings("back_in", i));
+            }
+        }
+
+        for (int i = 0; i < num_calls; ++i) {
+            instance.send("front_out", result[i], i);
         }
     }
 }
--- a/docs/source/examples/fortran/load_balancer.f90
+++ b/docs/source/examples/fortran/load_balancer.f90
@@ -17,7 +17,8 @@ program load_balancer
 
     type(LIBMUSCLE_Message) :: rmsg
     type(LIBMUSCLE_Message), allocatable :: result(:)
-    type(LIBMUSCLE_DataConstRef) :: rdata
+    type(LIBMUSCLE_Data) :: rdata
+    type(YMMSL_Settings) :: settings
 
     integer :: started, done, num_calls, num_workers, i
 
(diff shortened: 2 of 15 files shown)


src/cpp/libmuscle/port_manager.hpp
```diff
@@ -24,6 +25,8 @@ namespace libmuscle { namespace _MUSCLE_IMPL_NS {
 class PortManager {
     public:
         using PortMessageCounts = std::unordered_map<std::string, std::vector<int>>;
+        /* Memory safety: PortReferences is only valid while PortManager is alive! */
+        using PortReferences = std::vector<std::reference_wrapper<const Port>>;
```
LourensVeen, line 29: I'd say this works. The only alternative I can see is to make `ports_` a map from string to `std::shared_ptr<Port>` and then return a vector of shared pointers instead of of `std::reference_wrapper`. But that would signal shared ownership, while it's really `PortManager` that owns them. The way this worked, with copies being returned, was simply wrong given that the size can change while we're running. So let's do this.

docs/source/examples/python/reaction_diffusion_qmc.py
```diff
@@ -134,22 +134,26 @@ def load_balancer() -> None:
...
+        num_workers = instance.get_port_length("back_out")
+
+        instance.set_port_length("front_out", num_calls)
+        while len(result) < num_calls:
+            for i in range(num_workers):
+                if started < num_calls:
+                    msg = instance.receive_with_settings("front_in", started)
+                else:
+                    msg = Message(float("-inf"))  # data=None indicates nothing to do
+                instance.send("back_out", msg, i)
```
LourensVeen: So, are we sending a `None` message into muscle_settings_in here? How does that work? Won't that crash?

docs/source/examples/python/interact_coupling.py
```diff
@@ -317,38 +317,47 @@ def checkpointing_temporal_coupler() -> None:
...
+      right:
+        description: The other interacting component
+        ports:
+          o_i: boundary_out
+          s: boundary_in
+        implementation: model
+      coupler:
+        description: This connects the two and interpolates
+        ports:
+          timeline left:
```
LourensVeen, line 343: Hmm, should this be `timeline :left:`? Like this it would be `:coupler:left` I think, and then how does that match up with `:left` on the `left` component?

docs/source/examples/cpp/load_balancer.cpp
```diff
@@ -34,22 +34,28 @@ void load_balancer(int argc, char * argv[]) {
...
 
-        while (done < num_calls) {
-            while ((started - done < num_workers) && (started < num_calls)) {
-                auto msg = instance.receive_with_settings("front_in", started);
-                instance.send("back_out", msg, started % num_workers);
+        while (result.size() < num_calls) {
+            for (int i = 0; i < num_workers; ++i) {
+                auto msg = (started < num_calls) ? 
+                    instance.receive_with_settings("front_in", started)
+                    : Message(-INFINITY);
```
LourensVeen: `INFINITY` is C, and has moved to a different header in C23 apparently, so it's a bit iffy. In C++ you're supposed to use `std::numeric_limits<double>::infinity()` from the `<limits>` header. Also see comment on the Python version.

docs/source/examples/fortran/load_balancer.f90
```diff
@@ -36,22 +37,34 @@ program load_balancer
...
 
         do while (done < num_calls)
-            do while ((started - done < num_workers) .and. (started < num_calls))
-                rmsg = instance%receive_with_settings_on_slot('front_in', started)
-                call instance%send('back_out', rmsg, mod(started, num_workers))
+            do i = 0, num_workers-1
+                if (started < num_calls) then
+                    rmsg = instance%receive_with_settings_on_slot('front_in', started)
+                else
+                    rmsg = LIBMUSCLE_Message(-1d40)
```
LourensVeen: :rofl: Oh Fortran...

LourensVeen (commented):
This looks good, but I do have some questions. Thanks!

docs/source/examples/python/interact_coupling.py
maarten-ic: No, actually this is correct:
- Coupler itself lives in the root timeline (not on `:coupler`)
- This annotation indicates that `a_out` and `a_in` are in the subtimeline `left` of coupler, which resolves to `:left`
- The `left` component doesn't have a timeline annotation for their O_I and S ports. This means that they belong in the default subtimeline called after the name of that component. Since this component also lives on the root timeline, this resolves to the `:left` timeline for the `boundary_in` and `boundary_out` ports of `left`.

N.B. The following would be equivalent to the current yaml ports description for `left`:

```yaml
      left:
        description: One of the interacting components
        ports:
          timeline left:
            o_i: boundary_out
            s: boundary_in
```

I hope this clarification makes sense, let's discuss if not!

docs/source/examples/python/reaction_diffusion_qmc.py
maarten-ic: Good point... this works because the number of messages divides nicely and we never send that None message...

I guess this would be solved by sending an empty `Settings()` instead of `None` (and handling that situation downstream), I'll have a look into that.

docs/source/examples/cpp/load_balancer.cpp
maarten-ic: Good point, I'll update that!

docs/source/examples/python/interact_coupling.py
LourensVeen: Huh. I guess I don't understand how timelines work.

Now that I look at this, I think I once envisioned `timeline left` on `coupler` to put those ports into `:coupler.left`, with any O_I and S ports on `coupler` without an explicit timeline living in `:coupler`. Then you'd have to use `timeline :left` to put it into an absolute timeline, rather than one relative to the component.

But that's not how it now works and I have no idea whether that is because we decided to do it differently, or if it fell by the wayside without me noticing. I'll check my notes. Meanwhile, would love to have your opinion on which makes more sense in your view?

LourensVeen (commented):
Ouch, well, I guess we have an incentive to do something with vector timelines because that skip setting isn't pretty. But I don't see a better solution in the short term either, so let's do it like this. And let's discuss the timeline definition, see comment.

docs/source/examples/python/interact_coupling.py
maarten-ic: As discussed over Slack: this is a good point and we should discuss it in more detail.
In the meantime, this shouldn't block the work in MUSCLE3 for milestones and conduit filters.

LourensVeen: Yes, let's have a look at that separately, I don't think this will affect how things work, just how what's in the yMMSL file gets interpreted. So that should be changeable, and that will change the behaviour, but it shouldn't break anything.

LourensVeen (approved):
Okay, let's go with this and look at the semantics of the yMMSL annotation separately.

PR #394, opened 2026-08-13
Refactor send / receive

maarten-ic:
Some refactoring of send/receive logic in the communicator in preparation for the Milestones and Conduit Filter PRs:

1. Move "pre receive" of F_INIT messages to communicator, so the Instance won't need to care about conduit filters and the Communicator can handle it all.
2. Remove the "saved_until" metadata from MPPMessage and corresponding logic in the Instance/Communicator/TriggerManager. The checkpointing works just fine without it, and it simplifies the send/receive logic.
3. Slightly simplify constructing the endpoints: use `Optional[int]` for slots instead of a `list[int]` which may be `[]` or `[slot]`
4. Move some logic around so `send_message` and `receive_message` do things in the same order in the Python and C++ implementations.

### DIFF
--- a/src/cpp/libmuscle/communicator.cpp
+++ b/src/cpp/libmuscle/communicator.cpp
@@ -78,40 +78,33 @@ void Communicator::send_message(
         Message const & message,
         Optional<int> slot)
 {
-    std::vector<int> slot_list;
-    if (slot.is_set())
-        slot_list.push_back(slot.get());
-    log_debug("Sending message on ", port_desc(port_name, slot));
-
-    Endpoint snd_endpoint = get_endpoint_(port_name, slot_list);
-    if (!port_manager_.get_port(snd_endpoint.port).is_connected())
-        // log sending on disconnected port
-        return;
-
     Port & port = port_manager_.get_port(port_name);
-
-    ProfileEvent profile_event(
-            ProfileEventType::send, ProfileTimestamp(), {}, port, {}, slot,
-            port.get_num_messages(), {}, message.timestamp());
-
-    auto recv_endpoints = peer_info_.get().get_peer_endpoints(
-            snd_endpoint.port, slot_list);
-
-    Data settings_overlay(message.settings());
-
-    Optional<int> port_length;
-    if (port.is_resizable())
-        port_length = port.get_length();
+    if (!port.is_connected()) {
+        log_debug("Sending message on unconnected port ", port_desc(port_name, slot));
+        return;
+    }
+    log_debug("Sending message on ", port_desc(port_name, slot));
 
     Optional<IterationCount> iteration;
     if (!is_close_port(message.data()))
         iteration = timeline_manager_->check_send_message(port_name, slot);
 
+    ProfileEvent profile_event(
+            ProfileEventType::send, ProfileTimestamp(), {}, port, {}, slot,
+            port.get_num_messages(), {}, message.timestamp());
+
+    Optional<int> port_length;
+    if (port.is_resizable())
+        port_length = port.get_length();
+
+    auto endpoints = get_endpoints_(port, slot);
+    auto& snd_endpoint = std::get<0>(endpoints);
+    auto& recv_endpoints = std::get<1>(endpoints);
     for (auto recv_endpoint : recv_endpoints) {
         MPPMessage mpp_message(
                 snd_endpoint.ref(), recv_endpoint.ref(),
                 port_length, message.timestamp(), Optional<double>(),
-                settings_overlay, port.get_num_messages(slot),
+                Data(message.settings()), port.get_num_messages(slot),
                 message.data(), iteration);
 
         if (message.has_next_timestamp())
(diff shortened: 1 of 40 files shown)


src/python/libmuscle/test/conftest.py
```diff
@@ -115,7 +116,16 @@ def get_port(name):
     def port_exists(name):
         return name in mock_ports
 
+    def get_connected_ports(operator, timeline=None):
+        # Delegate to actual implementation
```
LourensVeen, line 120: I can see that it does that, but why?

src/python/libmuscle/timeline_manager.py
```diff
@@ -233,18 +233,10 @@ def __init__(self, port_manager: PortManager) -> None:
...
-        # do treat it as F_INIT though, so it must be tracked here as well.
-        if self._port_manager.settings_in_connected():
-            receive_ports.append(self._port_manager.get_port("muscle_settings_in"))
-        self._receive = TimelinePorts(receive_ports)
-
-        self._send = TimelinePorts(self._port_manager.get_connected_ports(Operator.O_F))
-
-        subtimelines = self._port_manager.list_subtimelines()
+        self._receive = TimelinePorts(port_manager.get_connected_ports(Operator.F_INIT))
+        self._send = TimelinePorts(port_manager.get_connected_ports(Operator.O_F))
```
LourensVeen, line 237: Nice!

src/python/libmuscle/instance.py
```diff
@@ -1154,32 +1150,32 @@ def _pre_receive(self) -> bool:
...
-        self.__pre_receive_f_init()
-        for message in self._f_init_cache.values():
-            if isinstance(message.data, ClosePort):
-                all_ports_open = False
-
-        if not all_ports_open:
+        try:
+            self._f_init_cache, saved_until = self._communicator.pre_receive_f_init()
+        except PortClosed:
             self._profiler.record_event(sw_event)
```
LourensVeen, line 1156: This modifies the meaning of `SHUTDOWN_WAIT` a bit because it now finishes when the first `ClosePort` is received rather than when all `F_INIT` messages have been received, but it arguably matches the definition in the docs more precisely now, so that's fine.

src/cpp/libmuscle/tests/test_communicator.cpp
```diff
@@ -231,6 +231,10 @@ TEST_F(libmuscle_communicator2, receive_close_port) {
     ASSERT_FALSE(mock_ports_.at("in")->is_open());
 }
 
+// Testing pre_receive_f_init is too difficult to set up correctly without the ability
```
LourensVeen, line 234: And that would take an extension to my horrible mocking template contraption? Can we make an issue for fixing this so we don't forget?

src/cpp/libmuscle/communicator.cpp
```diff
@@ -133,6 +133,42 @@ void Communicator::send_message(
         profiler_.record_event(std::move(profile_event));
 }
 
+std::tuple<Communicator::FInitCacheType, double> Communicator::pre_receive_f_init() {
+    FInitCacheType cache;
+    double max_saved_until = -std::numeric_limits<double>::infinity();
+
+    auto pre_receive = [&](std::string & port_name, Optional<int> slot) {
```
LourensVeen, line 140: Heh, I hadn't considered that you could make local functions using lambdas. Makes it more closely match the Python side and it's nice to have it all together.

src/python/libmuscle/checkpoint_triggers.py
```diff
@@ -190,19 +190,6 @@ def elapsed_walltime(self) -> float:
         """Returns elapsed wallclock_time in seconds."""
         return time.monotonic() + self._mono_to_elapsed
 
-    def checkpoints_considered_until(self) -> float:
-        """Return elapsed time of last should_save*"""
-        return self._cpts_considered_until
-
-    def harmonise_wall_time(self, at_least: float) -> None:
```
LourensVeen, line 197: So, this was never actually used? That's wild! How did that happen? My only worry is that we forgot to implement something when we made the checkpoints. We don't have any tests for clock skew, do we?

LourensVeen (approved):
All good, unless one of my questions makes you think it isn't :smile:.

src/python/libmuscle/test/conftest.py
maarten-ic: Fair, the comment doesn't add much indeed.
The tests sometimes set ports from connected=True to connected=False, so I wanted the actual implementation to do the filtering for me based on the current state of all the ports instead of hardcoding a return value.

src/python/libmuscle/instance.py
maarten-ic: I hadn't thought about that implication, but thankfully you did!

Note that with the upcoming Milestones there is no distinction: all F_INIT ports should be closed simultaneously or you didn't do timelines correctly

src/python/libmuscle/checkpoint_triggers.py
maarten-ic: Checkpointing isn't used a lot (in my experience) and checkpoints based on walltime are best-effort by design.

This was a pretty basic way of handling potential time skew between nodes, but I don't think it's worth the complexity in Communicator for what it did. If we want to improve checkpointing, then I would prefer to look into leveraging the milestones/iteration counts mechanism -- I expect that will be way more robust 🙂

src/python/libmuscle/test/conftest.py
maarten-ic: N.B. I'll update this comment in the Milestones PR (where I'm working on tests anyway) to make it more meaningful

Issue #395, opened 2026-08-16
Deadlock detector does not take into account waiters outside of the loop


LourensVeen:
The deadlock detection algorithm in the deadlock detector currently assumes that all deadlocks are waiting loops of the form `a -> b -> c -> a`. However, it's possible for other components to be waiting for something in that loop to send them data, e.g. `a -> b -> c -> b` where the deadlock is between `b` and `c` but `a` is waiting too.

In that second situation, when `a` asks if it's deadlocked, the manager goes into an infinite loop that slowly eats up all your RAM, because it keeps going around the `b -> c` loop looking for `a`.

A second problem is that when an instance reports that it's done waiting, only one potential deadlock is removed from the list. In the situation above, it can detect two potential deadlocks: `b -> c -> b` and `a -> b -> c -> b`. If `b` reports that it's continuing, then both need to be removed, but currently only the first one is, which later leads to a crash when trying to format a deadlock that no longer exists.

PR #396, opened 2026-08-16
Detect deadlocks with a tail

LourensVeen:
This updates the deadlock detector to be able to detect deadlocks that have a tail, i.e. where there's a loop of instances waiting for each other, but also some instances that aren't part of the loop but are waiting directly or indirectly for one of the instances in the loop.

I couldn't figure out an easy way of making that work with the existing design, where deadlocks are detected when a report comes in that something is waiting, and then later checked, because a report may come in that links the start of the tail to the loop and then you have to basically check all possible starting points every time. It seemed easier to just record the status of the instances and then try to create a cycle on the fly whenever we're asked whether there's a deadlock.

That seems to work, but maybe I missed something? Was there a performance concern maybe?

Addresses #395.

### DIFF
--- a/src/python/libmuscle/manager/deadlock_detector.py
+++ b/src/python/libmuscle/manager/deadlock_detector.py
@@ -24,8 +24,6 @@ class DeadlockDetector:
         """Maps instance IDs to the peer instance IDs they are waiting for."""
         self._waiting_instance_ports: dict[str, tuple[str, Optional[int]]] = {}
         """Maps instance IDs to the port/slot they are waiting for.."""
-        self._detected_deadlocks: list[list[str]] = []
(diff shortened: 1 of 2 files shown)


maarten-ic (approved):
Thanks! I like this updated design, much simpler!

> That seems to work, but maybe I missed something? Was there a performance concern maybe?

I don't remember why I had implemented it like this two years ago. Looking back at that PR there's been some iterative design of this logic and I missed that this could be simpler and more correct?

For performance: instances do one `waiting_for_receive()` and then 0+ times `is_deadlocked()`. This is a fast algorithm nevertheless so I don't see a problem with the new implementation (and in many cases we might end up doing less work because `waiting_for_receive()` is simpler now).

PR #397, opened 2026-08-18
Add Milestones

### LINKED ISSUE #382: Add Milestones for timelines
Placeholder issue for adding Milestone messages when an iteration is finished, this is a prerequisite for #374

### LINKED ISSUE #399: Add C++ communicator unit tests for `pre_receive_f_init`
And that would take an extension to my horrible mocking template contraption? Can we make an issue for fixing this so we don't forget?

_Originally posted by @LourensVeen in https://github.com/multiscale/muscle3/pull/394#discussion_r3806273800_

maarten-ic:
This Pull Request adds timeline Milestones to libmuscle.

Builds on top of #394, which should be merged first.

**Summary of changes**
1. Sending a message on a closed port now generates a `RuntimeError`
2. Removed `ClosePort` messages. A Milestone for the root timeline `Milestone([])` (a.k.a. "final milestone") is used to close ports now.
3. Each instance generates a Milestone at the end of their reuse loop. This Milestone is sent to O_I ports only because peers that are connected to O_F ports already know that they only get a single O_F message per reuse loop. This also ensures that the same Milestone messages are received for by `micro2` in a `macro -> micro2` connection as for a `macro -> micro1 -> micro2` connection.
4. Milestones that are received on an F_INIT port therefore originate from "higher" timelines than the one the component is in. These Milestones are broadcasted on all outgoing ports because this information is relevant for peer components in subtimelines (O_I connected) as well as those in our timeline (O_F connected).
5. Milestones that are received on an S port are currently ignored by the instance (unless it's a port closing final milestone). We could consider checking if the iteration counts match with what the TimelineManager expects for this subtimeline, but I have not implemented that now.

**Example**
image

- Component macro is in the root timeline. It will generate a Milestone for this timeline (`[]`) at the end of its reuse loop, this will also close all of its outgoing ports.
- Component meso is in the `:macro` timeline. It will generate Milestones for this timeline (`[0]`, `[1]`, ...) at the end of its reuse loop, which it sends on to micro (O_I port), but not back to macro (O_F port). When `macro` is done, `meso` receives the root milestone on its F_INIT port and broadcasts it to all outgoing ports.
- Component micro is in the `:macro:meso` timeline. It will generate Milestones for this timeline (`[0,0]`, `[0,1]`, ..., `[1,0]`, `[1,1]`, ...) at the end of its reuse loop and would send those on its O_I port if it had any. It will receive all milestones sent by meso on its F_INIT port and broadcasts those.


src/python/libmuscle/communicator.py
```diff
@@ -266,6 +292,36 @@ def pre_receive(port_name: str, slot: Optional[int]) -> None:
...
+        # Check if we have received milestones
+        milestone_iterations: list[Optional[IterationCount]] = []
+        closed_ports: set[str] = set()
+        for (port_name, _), message in cache.items():
+            it = message.data.iteration if isinstance(message.data, Milestone) else None
+            if it not in milestone_iterations:
+                milestone_iterations.append(it)
+            if it == []:
+                closed_ports.add(port_name)
+        # Milestones should be consistent, or we have a problem
```
LourensVeen: If we have a macro-meso-micro with an initialisation message from the root timeline going into micro, then micro is going to receive D[], M[] on the init port and D[0, 0], D[0, 1], M[0], D[1, 0], D[1, 1], M[1], M[] on the port from meso, and trigger this when it receives M[0] and M[1] even though that's perfectly legit.

I think just adding the iteration to `milestone_iterations` only if the message is a milestone, rather than also adding `None` if any data messages were received, should fix that.

src/python/libmuscle/communicator.py
```diff
@@ -266,6 +292,36 @@ def pre_receive(port_name: str, slot: Optional[int]) -> None:
...
+                raise RuntimeError(
+                    "Some ports were unexpectedly closed while trying to receive, did "
+                    "the peers crash? Closed ports: " + ", ".join(sorted(closed_ports))
+                )
+            raise RuntimeError(
+                "Internal error: received milestones for different iterations in "
+                f"F_INIT: {milestone_iterations}. This is not supposed to happen and "
+                "may be a bug in libmuscle. Please report an issue."
+            )
+        if milestone_iterations[0] is not None:
```
LourensVeen, line 294: This should then check for an empty list, and anyway who's to say that that `None` would be in the first position in the list? That depends on in which order we receive from the ports.

src/python/libmuscle/communicator.py
```diff
@@ -266,6 +292,36 @@ def pre_receive(port_name: str, slot: Optional[int]) -> None:
...
+            # Propagate milestone
+            milestone = Milestone(milestone_iterations[0])
+            self.broadcast_milestone(milestone, False)
+            if milestone.is_final_milestone():
```
LourensVeen, line 298: We've already raised `PortClosed` in `pre_receive` on line 280, so we'd never get here I think.

src/python/libmuscle/communicator.py
```diff
@@ -293,6 +345,20 @@ def receive_message(self, port_name: str, slot: Optional[int] = None) -> Message
             RuntimeError: If the network connection had an error, or the
                     message number was incorrect.
         """
+        # This method is only called by Instance to receive S messages. Instance should
```
LourensVeen: Maybe we can call this `receive_s_message` then to avoid confusion in the future?

src/cpp/libmuscle/milestone.cpp
```diff
@@ -0,0 +1,68 @@
...
+        throw std::runtime_error("Invalid milestone format. Bug in MUSCLE3?");
+
+    if (!obj_cache_)
+        obj_cache_ = std::make_shared<DataConstRef>(
+                mcp::unpack_data(mp_zones_->at(0), ext.data(), ext.size()));
+
+    return *obj_cache_;
+}
+
+bool is_milestone(DataConstRef const & data) {
```
LourensVeen, line 61: Should be declared in milestone.hpp?

src/cpp/libmuscle/milestone.hpp
```diff
@@ -0,0 +1,44 @@
...
+#include <libmuscle/data.hpp>
+#include <libmuscle/namespace.hpp>
+#include <libmuscle/timeline_manager.hpp>
+
+
+namespace libmuscle { namespace _MUSCLE_IMPL_NS {
+
+/** Represents a Milestone message.
+ *
+ * We need to be able to send a Milestone message just like we send user data
```
LourensVeen, line 12: Agreed.

src/cpp/libmuscle/tests/test_milestone.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+    msgpack::sbuffer buf;
+    msgpack::pack(buf, DataConstRef(milestone));
+
+    auto zone = std::make_shared<msgpack::zone>();
+    Data d2(unpack_data(zone, buf.data(), buf.size()));
+
+    ASSERT_TRUE(is_milestone(d2));
+
+    Milestone const * m2 = reinterpret_cast<Milestone const*>(&d2);
```
LourensVeen: Could this be a `dynamic_cast<>`?

src/cpp/libmuscle/tests/test_milestone.cpp
```diff
@@ -0,0 +1,81 @@
...
```
LourensVeen: I guess in real code we'd use `Milestone m2(d2)` instead of the cast? Then maybe we should test it here like that too?

LourensVeen (changes requested):
I reviewed commit-by-commit, so I hope this makes sense. It seems like maybe you fixed some of the issues I raised about the milestone receiving logic when you added the C++ side, but I'm not sure. Please check.

src/python/libmuscle/communicator.py
maarten-ic: Correct, but then we'll also have a repeater filter on the conduit from macro to micro -> these will need to be handled separately from the "unfiltered" conduits (to be implemented in the next PR 😉).

The algorithm I have in mind for that is:

F_INIT must have at least one connected port without repeater filters (this is checked by `ymmsl` in `resolve_timelines()`).
In macro->meso->micro with a direct connection macro-<repeat>->micro that would receive:

| macro->micro | meso->micro                     |
|--------------|---------------------------------|
| D[0]         | D[0,0]<br>D[0,1]<br>...<br>M[0] |
| D[1]         | D[1,0]<br>D[1,1]<br>...<br>M[1] |
| ...          | ...                             |
| M[]          | M[]                             |

The milestones on the unfiltered conduit(s) indicate that we can receive another message on the repeated conduits.

We still need this Milestone consistency check for all conduits without repeater filters and we should get an error when a message on one F_INIT port is mixed with a milestone on another one.
As an example, I'm taking the following message order on port A: `D[0,0], D[0,1], D[0,2], M[0]`.
Then for another port B also connected on F_INIT we expect the exact same iteration counts and milestones in the exact same order. The following would be wrong:

1. `D[0,0], D[0,1], M[0]` -> We should get message for D[0,2] as well on this port, otherwise the incoming timelines are not consistent
2. `D[0,0], D[0,1], D[0,2], D[1,0]` -> Missing milestone message, that's a problem as well.
3. `D[0,0], D[0,1], D[0,2], D[0,3], M[0]` -> Inconsistent timelines as in case 1, but now one message more compared to port A

I hope this makes sense, we can set up a call to discuss if you like!

src/python/libmuscle/communicator.py
maarten-ic: At this point in the logic, this list will be of size 1: it is not >1 (this is checked on line 305) and `pre_receive` is skipped when there are no connected F_INIT ports so it cannot be empty either.

src/python/libmuscle/communicator.py
maarten-ic: Indeed! Found and fixed this in the second commit, still thanks for the note!

src/python/libmuscle/communicator.py
maarten-ic: Good suggestion! Will do

src/cpp/libmuscle/milestone.cpp
maarten-ic: It needs to be a friend of DataConstRef, so it's declared in `data.hpp`. It is a bit obscure, but this was also how it was done for `is_close_port` previously: I just renamed the declaration.

src/cpp/libmuscle/milestone.hpp
maarten-ic: Glad you do! I just copied and slightly updated what you wrote for the ClosePort class that Milestone is replacing 😉

src/cpp/libmuscle/tests/test_milestone.cpp
maarten-ic: Ah, it should just be a `Milestone(d2)`. I previously had `reinterpret_casts` everywhere but was very unhappy with that pattern. Apparently I forgot to change it in these tests -> will update!

src/python/libmuscle/communicator.py
LourensVeen: Ah, I missed that `pre_receive` is never called without F_INIT ports, but this was also a follow-on to the previous comment about not adding anything to `milestone_iterations` for ports on which we receive a non-milestone message. In that case you can have an empty list if there are no milestones.

src/python/libmuscle/communicator.py
LourensVeen: That does make sense, and all good implementing that in the next PR.

src/cpp/libmuscle/milestone.cpp
LourensVeen: Ah, I guess it made sense when I wrote it :smile:. Yeah it's really kind of a member function of DCR, but you don't want to have it there, so it's a separate function, and I probably didn't want to have it in `data.hpp` because that's part of the public API.

src/cpp/libmuscle/milestone.hpp
LourensVeen: Haha! I didn't look at the deleted files, so I missed that. But glad I'm consistent at least :sweat_smile:

LourensVeen (approved):
I'm happy with this, let's go!

Issue #164, opened 2023-01-26
Sending on a vector port should require a slot


LourensVeen:
Right now if you omit the slot, it'll send on slot 0, while we want an error message explaining that something went wrong.

PR #398, opened 2026-08-18
Refactor send/receive error messages

maarten-ic:
- Clean up logic for default messages when receiving on disconnected ports
- Better alignment of Python and C++ error messages
- Add a check that a slot is given when sending or receiving on a vector port (see #164)

N.B. Builds on top of #397, which should be merged first.


LourensVeen (approved):
Less code more better. Excellent!

Issue #294, opened 2024-05-04
Feature request: inclusion of environment variables in the YMMSL files


DavidPCoster:
It would be very useful if one could use environment variables in the YMMSL files.

Suggestion for format: ${ENVIRONMENT_VARIABLE_NAME}

I think it would make sense to do this in the environment of the muscle_manager interpreting the YMMSL files.

The configuration.log would show the instantiated values.

LourensVeen:
I'm going to close this as a duplicate of https://github.com/multiscale/muscle3/issues/246. That one got a bit messy, but the original request is the same, and the things in there that aren't environment variables in YMMSL files now have a solution.

Issue #246, opened 2023-06-27
Would it be possible to get environment variables in the YMMSL file?[/-]


DavidPCoster:
Would it be possible to get environment variables in the YMMSL file?

It would be good to be able to have something like
  work_dir: ${HOME}/GIT/ets_paf/UQ
in the YMMSL file.

LourensVeen:
Ah, work dir. You're not the only one to want to be able to run in a work dir that's not inside the run dir. It's currently not really supported because it's problematic from a provenance and reproducibility perspective, you end up overwriting previous results, or possibly depending on them.

You can hack your way around it by using `cd ${HOME}/GIT/ets_paf/UQ ; <command>` as the command, as the programs are started inside of a shell. That also allows you to use environment variables.

There may be some changes in the future however with exactly how things are started, currently it's a login shell but that's not ideal. I think it may be best to make it configurable, but I need some support from QCG-PJ for that. There's #201 for this.

DavidPCoster:
Going into the worker's would be fine -- however I seem to be in my home directory instead ...

DavidPCoster:
OK -- I found a cd lurking in a system file sourced by the local .bash_profile

After removing that, the workdir functionality works.

I would still like to be able to include environment variables in the YMMSL file, though ...

LourensVeen:
Well, that should be possible. From which environment should they be taken though? The one the `muscle_manager` is started from? A clean (login) shell on the compute node? The laptop from which you ran `ssh -c sbatch ...`? And if the environment variable is used in the settings, how does that work with dynamically generated settings and `muscle_settings_in`? Or should this be limited to the implementations?

I'm trying to figure out the usage scenarios that people may have, because I think it would be easy to make something that doesn't do what at least some users expect. Having an executable location relative to `${HOME}` makes sense, especially on clusters where the home dir is often not in `/home/user` or may be mounted in a different place on the head node and a compute node. Are there other things you would like to pull from the environment?

DavidPCoster:
I suggest that the environment variables be those seen by muscle_manager ...

AreWeDreaming:
I second this and would also like to have the ability to use environment variables. For example for the path to the applications. Hardcoding those is quite nasty.

AreWeDreaming:
[...]

PR #400, opened 2026-08-19
Expand environment variables in settings

Yannicked:
This PR adds the ability to expand environment variables inside settings

### DIFF
--- a/src/python/muscle3/muscle_manager.py
+++ b/src/python/muscle3/muscle_manager.py
@@ -1,5 +1,4 @@
 import os
-import re
 import sys
 import textwrap
 import traceback
@@ -228,9 +227,6 @@ def _manage_simulation(
             print(f"Output may be found in {run_dir_obj.path}")
 
     sys.exit(0 if success else 1)
-
-
-_ENV_VAR_RE = re.compile(r"\$(\w+|\{[^}]*\})")
 
 
 def expand_settings(config: v0_2.Configuration) -> v0_2.Configuration:
--- a/src/python/muscle3/muscle_manager.py
+++ b/src/python/muscle3/muscle_manager.py
@@ -1,3 +1,5 @@
+import os
+import re
 import sys
 import textwrap
 import traceback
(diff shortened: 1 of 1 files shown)


maarten-ic:
We should discuss how to approach this and what we want this feature to do. See also #294 for a related issue.

Some high-level thoughts:
1. I think this feature should live in ymmsl-python, so all tools reading ymmsl can use the same functionality
2. This feature may break existing configurations that use `$name` or `${name}` in their settings
3. `os.path.expandvars()` doesn't appear to allow "escaping", so if I really want a setting containing "$HOME" then there's no way to achieve that:
```python
>>> os.path.expandvars("\\$HOME")
'\\/home/maarten'
>>> os.path.expandvars("$$HOME")
'$/home/maarten'
```
4. It may be confusing for users if we only do environment variable expansion of settings, and not in other useful places like executables

LourensVeen:
Whoa, yes, thanks, but let's discuss this first, because nothing is ever simple in MUSCLE3 :smile:. Here's two points that I have to add. No, three. Four! Sorry for the long comment.

First, settings aren't static, but can be changed during execution by sending a message to `muscle_settings_in`. So if we have a setting `inputfile: something` but it gets overridden with `inputfile: $HOME/something` then we need to substitute the environment variable in libmuscle and after each pre-receive, not in ymmsl. Although I suppose libmuscle could call a function in ymmsl-python to actually do it, but only if it's the Python libmuscle, the C++ libmuscle would have to do it some other way.

Second, MUSCLE2 didn't have yMMSL (although it had xMML, but I don't think that was really used). Instead, it used a Ruby script that created a set of objects describing the coupling. I don't remember how that got to the Java side, maybe the Ruby script called a function that started everything, but anyway, an advantage of that was that you could define settings as a function of another setting, e.g.:
```
r_toroidal = 8.0
r_poloidal = 3.0
diameter = 2.0 * (r_toroidal + r_poloidal)
```
I've always thought that it would be neat if we could do that in MUSCLE3 as well by embedding expressions into the settings (especially useful if one of the variables gets varied dynamically in a parameter sweep or UQ run), and that expression language could then have a function for getting environment variables.

The tricky thing is that because of the above you need an expression parser that can be called from both Python and C++. I've looked at that years ago and found some header-only expression parser libraries that looked like they could be wrapped in Python quite easily, but I never got around to implementing it.

Third, it seems that many of the use cases for substitution of environment variables are related to staging in input files. We don't currently have a facility for this, but workflow systems commonly do. If that's the actual requirement then it may be better to solve that problem, rather than making it easier to work around. On the other hand, it's also pretty easy to write a wrapper script around the executable that sets up the run directory, so it's not urgent.

Fourth, there's the question of how to escape things. YAML has quite a few reserved characters, and likes to automatically recognise all sorts of stuff in some circumstances but not others. The `$` character is actually one of the few that's not used for something, so that helps. We could have `${NAME}` for environment variables while leaving`$(2.0 * (r_toroidal + r_poloidal))` for a future expressions feature. As Maarten pointed out, we need some way to escape that because we may well have a model that takes input with funny characters like `$`.

So, question for @Yannicked: which problem does this solve, and is it the best solution for that?

LourensVeen:
Discussion continued in the OG issue #246.

PR #401, opened 2026-08-21
Add ConduitFilters to the muscle manager protocol

maarten-ic:
- Add ConduitFilters to C++ ymmsl API
- Muscle Manager now sends the configured conduit filters for each Conduit to the instances.
- Instances store the conduit filters in PeerInfo

### DIFF
--- a/src/cpp/ymmsl/model.hpp
+++ b/src/cpp/ymmsl/model.hpp
@@ -7,6 +7,39 @@
 
 namespace ymmsl { namespace impl {
 
+/** Represents a filter to apply to messages passing through a conduit.
+ */
+enum ConduitFilter {
+    /** Pass only the last message and drop preceding ones. */
+    LAST,
+    /** Repeat a single message as often as needed. */
+    REPEAT,
+    /** Pass a single message and then generate nil-messages as needed. */
+    PAD
+};
+
+/** Return the name of the given conduit filter.
+ */
+std::string conduit_filter_name(ConduitFilter filter);
+
+/** Return the conduit filter from its name.
+ */
+ConduitFilter conduit_filter_for_name(std::string const & name);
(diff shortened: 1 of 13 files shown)


IrisvdWerf (approved):

src/python/libmuscle/peer_info.py
```diff
@@ -72,30 +69,6 @@ def list_ymmsl_ports(self) -> list[Port]:
         """list ports declared in the yMMSL configuration"""
         return self._ymmsl_ports
 
-    def list_incoming_ports(self) -> list[tuple[Identifier, Reference]]:
```
LourensVeen, line 75: I guess an earlier change made these dead code?

src/cpp/ymmsl/model.hpp
```diff
@@ -7,6 +7,39 @@
 
 namespace ymmsl { namespace impl {
 
+/** Represents a filter to apply to messages passing through a conduit.
+ */
+enum ConduitFilter {
```
LourensVeen: Can we make that an enum class? That's a bit cleaner than the old style ones.

LourensVeen (changes requested):
These smaller PRs are easier to review, thanks! This looks good, but I have a small request on the enum class.

src/python/libmuscle/peer_info.py
maarten-ic: I suppose so as well, but since it's not used there's no need to keep it around 😄

src/cpp/ymmsl/model.hpp
maarten-ic: Of course!

LourensVeen (approved):
Very classy :smile:

PR #402, opened 2026-08-24
Implement repeater conduit filters

maarten-ic:
This PR adds support for "repeater" conduit filters connected to F_INIT ports.

This updates the logic for pre-receiving F_INIT messages to:
1. Receive op unfiltered F_INIT ports
2. Check received messages for consistency of Milestones and Iteration counts
3. If milestone received:
   1. Check if cached repeated messages are still valid, discard those that are too old
   2. Go back to 1
4. For each repeated F_INIT port
   1. If there is no message in the cache, we need to receive it from the peer. This is a while loop because we may need to receive any number of messages (see comment and associated test case).
   2. Copy the message from the cache
   3. Set data to nil if we need to pad rather than repeat

This PR also removes the F_INIT receive checks in the TimelineManager. The pre-receive logic in communicator is responsible for consistency now.

N.B.
1. I did not add support for repeated messages on S ports. In principle this is possible to add, but it would further complicate the implementation. Users are now instructed to connect to an F_INIT port instead. @LourensVeen Do you know of a use case where it makes sense to connect a conduit `output -> repeat -> o_i port` instead of `output -> f_init port`?
2. The F_INIT message cache should be saved and restored when making checkpoints. I will add this in the following PR to keep the scope of this one somewhat manageable.

### DIFF
--- a/integration_test/test_filters.py
+++ b/integration_test/test_filters.py
@@ -1,0 +1,127 @@
+import pytest
+from ymmsl.v0_2 import Operator
+
+from libmuscle import Instance, Message
+
+from .conftest import run_manager_with_actors
+
+
+def macro() -> None:
+    instance = Instance({Operator.O_I: ["out"]})
+
+    while instance.reuse_instance():
+        for i in range(4):
+            instance.send("out", Message(i, data=["macro", i]))
+
+
+def meso() -> None:
+    instance = Instance({Operator.F_INIT: ["in"], Operator.O_I: ["out"]})
+
+    reused = 0
+    while instance.reuse_instance():
+        msg = instance.receive("in")
+        for i in range(reused):
+            instance.send("out", Message(i, data=msg.data + ["meso", i]))
+        reused += 1
+
+
+def micro() -> None:
+    instance = Instance({Operator.F_INIT: ["in"], Operator.O_I: ["out"]})
+
+    while instance.reuse_instance():
+        msg = instance.receive("in")
+        for i in range(2):
+            instance.send("out", Message(i, data=msg.data + ["micro", i]))
+
+
+def pico(filters: str) -> None:
+    print("Running with conduit filter:", filters)
+    instance = Instance({Operator.F_INIT: ["macro", "meso", "micro"]})
+
+    reused = 0
+    expected_counts = [
+        [1, 0, 0],
+        [1, 0, 1],
+        [2, 0, 0],
+        [2, 0, 1],
+        [2, 1, 0],
+        [2, 1, 1],
+        [3, 0, 0],
+        [3, 0, 1],
+        [3, 1, 0],
+        [3, 1, 1],
+        [3, 2, 0],
+        [3, 2, 1],
+    ]
+    while instance.reuse_instance():
+        macro = instance.receive("macro")
+        meso = instance.receive("meso")
+        micro = instance.receive("micro")
+
+        print("Received:", macro.data, meso.data, micro.data)
+        message_counts = micro.data[1::2]
+        assert message_counts == expected_counts[reused]
+
+        # Test if macro message is padded or repeated, based on conduit filter filters
+        if filters == "pad pad" and any(message_counts[1:]):
+            assert macro.data is None
+        elif filters == "repeat pad" and message_counts[-1]:
+            assert macro.data is None
+        else:
+            assert macro.data == micro.data[:2]
+
+        assert micro.data[:4] == meso.data
+
+        assert micro.data[1::2] == expected_counts[reused]
+        reused += 1
+
+    assert reused == len(expected_counts)
+
+
+config = """
+ymmsl_version: v0.2
+models:
+  repeaters:
+    components:
+      macro:
+        description: macro
+        ports:
+          o_i: out
+        implementation: macro
+      meso:
+        description: meso
+        ports:
+          f_init: in
+          o_i: out
+        implementation: meso
+      micro:
+        description: micro
+        ports:
+          f_init: in
+          o_i: out
+        implementation: micro
+      pico:
+        description: pico
+        ports:
+          f_init: macro meso micro
+        implementation: pico
+    conduits:
+      macro.out:
+      - meso.in
+      - {filters} pico.macro
+      meso.out:
+      - micro.in
+      - repeat pico.meso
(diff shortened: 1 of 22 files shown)


LourensVeen:
> @LourensVeen Do you know of a use case where it makes sense to connect a conduit output -> repeat -> s port instead of output -> f_init port?

I can imagine a model that expects to receive a boundary condition, say a flux at the edge of its domain, on every time step, and then wanting to impose a fixed boundary condition by sending the same message on every time step. A natural way to do that would be to have a component read or calculate that data and send it on O_F, then connecting that to the S port with a repeat filter.

Connecting from a micro O_F to a macro S bypassing the meso in the middle is another case of this. So I think only allowing this on F_INIT is only half the solution.

src/python/libmuscle/timeline_manager.py
```diff
@@ -247,6 +252,28 @@ def __init__(self, port_manager: PortManager) -> None:
             None if self._port_manager.has_f_init_connections() else []
         )
 
+    def check_f_init_iterations(
+        self, iterations: dict[PortAndSlot, IterationCount]
+    ) -> IterationCount:
+        """Set current iteration count from pre-received F_INIT messages."""
+        assert self._iteration is None
+        assert iterations
+        first, *remainder = iterations.values()
```
LourensVeen, line 261: 😮 Wait, you can do that? What is this, Haskell? Cool!

src/python/libmuscle/timeline_manager.py
```diff
@@ -247,6 +252,28 @@ def __init__(self, port_manager: PortManager) -> None:
             None if self._port_manager.has_f_init_connections() else []
         )
 
+    def check_f_init_iterations(
+        self, iterations: dict[PortAndSlot, IterationCount]
```
LourensVeen, line 256: I  like this type.

src/python/libmuscle/timeline_manager.py
```diff
@@ -247,6 +252,28 @@ def __init__(self, port_manager: PortManager) -> None:
...
+    ) -> IterationCount:
+        """Set current iteration count from pre-received F_INIT messages."""
+        assert self._iteration is None
+        assert iterations
+        first, *remainder = iterations.values()
+        for it in remainder:
+            if it != first:
+                raise RuntimeError(
+                    "Pre-received F_INIT messages from parallel timelines:\n"
```
LourensVeen, line 265: What are "parallel timelines" here? I would interpret that maybe as e.g. two macro models in lockstep calling the same micro model, but then that would actually pass here. Isn't this about "different timelines"?

src/python/libmuscle/timeline_manager.py
```diff
@@ -247,6 +252,28 @@ def __init__(self, port_manager: PortManager) -> None:
...
+                    + "\n".join(
+                        f"- {port_desc(port, slot)}: {it}"
+                        for (port, slot), it in iterations.items()
+                    )
+                    + "\nPlease verify that all connected instances send the same "
```
LourensVeen, line 270: This could get complicated if there's e.g. a filter in between. Maybe it would be better to say "verify that all F_INIT ports receive the same number of messages"? Or "receive message simultaneously"?

maarten-ic:
> > @LourensVeen Do you know of a use case where it makes sense to connect a conduit output -> repeat -> s port instead of output -> f_init port?
> 
> I can imagine a model that expects to receive a boundary condition, say a flux at the edge of its domain, on every time step, and then wanting to impose a fixed boundary condition by sending the same message on every time step. A natural way to do that would be to have a component read or calculate that data and send it on O_F, then connecting that to the S port with a repeat filter.
> 
> Connecting from a micro O_F to a macro S bypassing the meso in the middle is another case of this. So I think only allowing this on F_INIT is only half the solution.

That makes sense, thanks! I'll implement repeaters for S filters as well in a follow-on PR

src/python/libmuscle/communicator.py
```diff
@@ -291,16 +355,56 @@ def pre_receive(port_name: str, slot: Optional[int]) -> None:
...
-            self._broadcast_milestone(milestone, False)
-            if milestone.is_final_milestone():
-                raise PortClosed()
-            # Pre-receive again to receive the actual messages
-            return self.pre_receive_f_init()
+        (iteration,) = milestone_iterations
+        return None if iteration is None else Milestone(iteration)
 
-        return cache
+    def _pre_receive_f_init_with_repeaters(
```
LourensVeen, line 361: The complexity here worries me. In my original (emailed) description of the timelines, I had a different algorithm, that does all the F_INIT ports together:

>     # on startup
>     cur_iteration = []
>     open_f_init_ports = {all connected f_init ports}
> 
>     # at pre-receive
>     milestones = {M[cur_iteration]}     # the iteration about to be replaced
>     while milestones not empty:
>         for each message in the cache:
>             if its iteration matches one of the milestones' iterations:
>                 delete it from the cache
> 
>         for each port in open_f_init ports:
>             if there's no message in the cache for it:
>                 receive a new message into the cache
> 
>         # send buffered messages triggered by any received milestones (see Reducers)
> 
>         # forward any received milestones (see Milestones above)
> 
>         open_f_init_ports -= all ports for which we have M([]) in the cache
>         milestones = move milestones from cache while deduplicating
> 
>     if open_f_init_ports is empty, and this is not the first run of a component with no
>     f_init ports, then shut down
> 
>     cur_iteration = longest iteration from all messages in the cache

That seems simpler than splitting it up, but it may expand with all the error handling and checks and stuff.

integration_test/test_filters.py
```diff
@@ -0,0 +1,139 @@
...
+def micro() -> None:
+    instance = Instance({Operator.F_INIT: ["in"], Operator.O_I: ["out"]})
+
+    while instance.reuse_instance():
+        msg = instance.receive("in")
+        for i in range(2):
+            instance.send("out", Message(i, data=msg.data + ["micro", i]))
+
+
+def pico(filters: str) -> None:
```
LourensVeen, line 37: :rofl:

LourensVeen (commented):
I've looked only at the Python side, and have some comments. Let's discuss what to do.

maarten-ic:
PR replaced by #410

Issue #379, opened 2026-07-09
Conduits from a component to itself cause the instance to crash


LourensVeen:
It's possible in MUSCLE3 to connect a conduit from a port on an instance to another port on that same instance, for example if you have a macro model calling a micro model that shares the state, and so you're sending the state back and forth, but now you want to run without the micro model and so you connect the ports together.

However, if you try this, the instance crashes, saying that the incoming port has two incoming conduits, which is illegal. This is caused by a bug in `TopologyStore.get_conduits`, which adds the conduit to the list twice, once because we're sending on it and once because we're receiving on it. This list ends up in `PeerInfo` in the instance, which then raises an error on line 57 when processing the second copy of the conduit.

Solution: in topology_store.py:47, turn `if` into `elif` so that the conduit only gets added once in this case.

PR #403, opened 2026-08-25
Fix self-connection conduit double count (#379)

LourensVeen:
This fixes the topology store to not double count conduits from a component to itself. Addresses #379.


maarten-ic (approved):
Looks good to me!

PR #406, opened 2026-08-27
Refactor decide reuse instance

maarten-ic:
Simplify things by:
- Ensuring that `Communicator.pre_receive_f_init()` is always called at the beginning of the reuse loop.
  - We will need this for repeater conduit filters on S ports.
  - No separate `finish_reuse_iteration` call anymore, this is now handled during the pre_receive
  - TimelineManager is now also notified at the start of the reuse loop (`finish_reuse_iteration` -> `start_reuse_iteration`)
- Removing `Instance.decide_reuse_instance` and move most logic (back) to reuse_instance().
  - `should_check_final_checkpoint` needs to know only if new F_INIT messages are pre-received: all logic related to resuming from a snapshot and whether this is the first run is only relevant in reuse_instance()
  - This makes it easier to add the relevant Communicator state (the message cache) to the snapshots later

### DIFF
--- a/src/python/libmuscle/communicator.py
+++ b/src/python/libmuscle/communicator.py
@@ -165,11 +165,6 @@ class Communicator:
         """
         self._timeline_manager.restore_state(state)
 
-    def finish_reuse_iteration(self) -> None:
-        """Finish the reuse iteration and send milestones."""
-        finished_iteration = self._timeline_manager.finish_reuse_iteration()
-        self._broadcast_milestone(Milestone(finished_iteration), True)
-
     def send_message(
         self,
         port_name: str,
@@ -254,6 +249,10 @@ class Communicator:
         Returns:
             f_init_cache: The received messages.
         """
+        finished_iteration = self._timeline_manager.start_reuse_iteration()
+        if finished_iteration is not None:
+            self._broadcast_milestone(Milestone(finished_iteration), True)
+
         cache: FInitCacheType = {}
 
         def pre_receive(port_name: str, slot: Optional[int]) -> None:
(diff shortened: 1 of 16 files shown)


src/python/libmuscle/timeline_manager.py
```diff
@@ -243,9 +243,7 @@ def __init__(self, port_manager: PortManager) -> None:
 
         # A component with no connected F_INIT ports has no F_INIT message to
         # learn its iteration from, so its main timeline starts at [].
-        self._iteration: Optional[IterationCount] = (
-            None if self._port_manager.has_f_init_connections() else []
-        )
+        self._iteration: Optional[IterationCount] = None
```
IrisvdWerf, line 246: What happens to a component with no F-INIT ports? How is its iteration set to []? Also, if check_send_message asserts when the value is None, does that mean we cannot send on an O_I or O_F port when nothing has been received?

src/python/libmuscle/communicator.py
```diff
@@ -254,6 +249,10 @@ def pre_receive_f_init(self) -> FInitCacheType:
         Returns:
             f_init_cache: The received messages.
         """
+        finished_iteration = self._timeline_manager.start_reuse_iteration()
```
IrisvdWerf, line 252: Feels a bit confusing that start_reuse_iteration returns a finished_iteration, although this is correct.

src/python/libmuscle/timeline_manager.py
maarten-ic: > How is its iteration set to []?

This happens in `start_reuse_iteration` when calling `TimelineManager.reset()`, but I think this will be updated with the upcoming conduit filters PR.

> does that mean we cannot send on an O_I or O_F port when nothing has been received?

If we have F_INIT ports, pre_receive_f_init will guarantee that we receive on it before the user will send on an O_I or O_F port. If we have no F_INIT ports, then the iteration is set to `[]` (see above).

So I think this works, right?

src/python/libmuscle/communicator.py
maarten-ic: I'm happy to change if it's too confusing. Do you have a suggestion to improve it?

LourensVeen (approved):
Okay, yes, that looks good. I got a bit confused about the logic at first and tried to simplify it and failed and now I understand :smile:.

IrisvdWerf (approved):

Issue #285, opened 2024-01-07
muscle_settings_in not drained


LourensVeen:
At least in the C++ version of libmuscle, it seems like we're not draining `muscle_settings_in` when closing incoming ports, so that we'll end up hanging if we have something trying to send us new settings while something else that we have an `F_INIT` port from has quit.

Should check if it's also the case in Python, and fix where needed. And add some regression tests.

Actually, maybe what we should do is make `muscle_settings_in` as much as possible a normal (although hidden from the user-facing API) port, rather than the special status it has now.

PR #407, opened 2026-08-27
Close muscle_settings_in on shutdown

maarten-ic:
Update Communicator.close_ports:

- Close `muscle_settings_in` port (if connected), which was never done previously
- Skip closing ports when not connected yet

N.B. builds on top of #406, which should be merged first.

### DIFF
--- a/src/python/libmuscle/communicator.py
+++ b/src/python/libmuscle/communicator.py
@@ -580,25 +580,24 @@ class Communicator:
         there will be no more messages, and allows the sending instance to shut down
         cleanly.
         """
-        for operator, port_names in self._port_manager.list_ports().items():
-            if operator.allows_receiving():
-                for port_name in port_names:
-                    port = self._port_manager.get_port(port_name)
-                    if not port.is_connected():
-                        continue
-                    try:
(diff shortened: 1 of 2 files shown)


LourensVeen:
Related: #285

src/python/libmuscle/communicator.py
```diff
@@ -580,31 +580,32 @@ def _close_incoming_ports(self) -> None:
...
-                        _logger.warning(
-                            "Connection with peer '%s' was lost at the end of the "
-                            "simulation, probably because it crashed.",
-                            peer_name,
-                        )
+        for operator in (Operator.F_INIT, Operator.S):
+            for port in self._port_manager.get_connected_ports(operator):
+                port_name = str(port.name)
+                if not port.is_connected():
+                    continue
```
IrisvdWerf: Loop over connected ports, so can this if-statement ever be true?

maarten-ic: Sharp, updated now!

IrisvdWerf (approved):

LourensVeen (approved):
I agree, nice!

PR #409, opened 2026-08-28
Refactor internal snapshot state

maarten-ic:
- Create CommunicatorState which captures both the `port_message_counts` and `timeline_state`. This will store the message cache (for repeated messages) as well in the future
- Store CommunicatorState before calling pre-receive, instead of trying to correct the state after the pre-receive

### DIFF
--- a/src/python/libmuscle/communicator.py
+++ b/src/python/libmuscle/communicator.py
@@ -3,6 +3,7 @@ from typing import Any, Optional, cast
 
 from ymmsl.v0_2 import Operator, Reference, Settings
 
+from libmuscle.communicator_state import CommunicatorState
 from libmuscle.endpoint import Endpoint
 from libmuscle.mcp.tcp_util import SocketClosed
 from libmuscle.mmp_client import MMPClient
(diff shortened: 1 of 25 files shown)


src/python/libmuscle/communicator.py
```diff
@@ -151,19 +152,23 @@ def set_receive_timeout(self, receive_timeout: float) -> None:
...
+            self._timeline_manager.get_state(),
+        )
 
-    def restore_state(self, state: TimelineState) -> None:
+    def restore_state(self, state: CommunicatorState) -> None:
         """Restore the timeline manager to a previously saved state.
 
         Args:
             state: The state to restore, as returned by get_state().
         """
```
IrisvdWerf, line 170: Update this docstring, as this also includes the message counts?

LourensVeen (approved):
I like the concept, and it looks cleaner, so nice! Also I agree with Iris about that comment, otherwise let's go :smile:

IrisvdWerf (approved):

PR #410, opened 2026-08-31
Implement repeat filters for F_INIT and S ports

maarten-ic:
Replaces #402 

This PR implements repeat/pad filters for F_INIT and S ports.

**Brief overview of changes:**
- Update communicator:
  - In addition to F_INIT messages, we also pre-receive S messages with repeat/pad filters applied.
  - All pre-received MPPMessages are stored in a new `message_cache`.
  - The message cache is saved and restored with the checkpointing mechanism.
- Update timeline manager
  - The timeline manager no longer verifies F_INIT receives. Communicator.pre_receive takes care of that now and calls TimelineManager.check_f_init_iteration_counts to verify the iteration counts on the messages (and update the current iteration of the component).
  - Additional logic to handle iteration counts for repeated S messages.
- Additional tests
  - Unit tests for the communicator are in a new file `test_communicator_conduit_filters.py`/`.cpp`: these tests use fewer mocked components (only MPPClient, MMPClient and Profiler) and test how the Communicator responds to mocked message streams.
  - Integration tests (`integration_test/test_filter.py`) for conduit filters, checking pad/repeat and testing the checkpointing mechanism.

### DIFF
--- a/src/cpp/libmuscle/communicator.cpp
+++ b/src/cpp/libmuscle/communicator.cpp
@@ -33,6 +33,45 @@ using ymmsl::Settings;
 
 namespace libmuscle { namespace _MUSCLE_IMPL_NS {
 
+namespace {
+
+/** Helper method to construct a Message from an MPPMessage.
+ * 
+ * N.B. since C++ enforces only const access to .settings and .data, we don't need to
+ * make a copy (unlike the Python equivalent).
+ */
+Message make_message(MPPMessage const & mpp_msg) {
+    Message message(
+        mpp_msg.timestamp,
+        mpp_msg.data,
+        mpp_msg.settings_overlay.as<Settings>()
+    );
+    if (mpp_msg.next_timestamp.is_set())
+        message.set_next_timestamp(mpp_msg.next_timestamp.get());
+    return message;
+}
+
+
+/** Helper template method to execute code for each slot of the given port.
+ * 
+ * Expects a function with arguments (Optional<int> slot, Reference port_ref)
+ */
+template<typename F>
+void for_each_slot(Port const & port, F&& f) {
+    Reference port_name(port.name);
+    if (!port.is_vector()) {
+        f({}, port_name);
+    } else {
+        int slot = 0; // Allow pre-receive to receive 1 message that updates port._length
+        do {
+            f(slot, port_name + slot);  
+        } while (++slot < port.get_length());
+    }
+}
+
+}
+
+
 Communicator::Communicator(
         ymmsl::Reference const & kernel,
         std::vector<int> const & index,
@@ -47,6 +86,9 @@ Communicator::Communicator(
     , server_()
     , clients_()
     , receive_timeout_(10.0)  // Notify manager, by default, after 10 seconds waiting in receive_message()
+    , pre_receive_ports_()
+    , repeat_filters_()
+    , message_cache_()
 {}
 
 std::vector<std::string> Communicator::get_locations() const {
@@ -56,13 +98,15 @@ std::vector<std::string> Communicator::get_locations() const {
 void Communicator::set_peer_info(PeerInfo const & peer_info) {
     peer_info_ = peer_info;
     timeline_manager_ = std::make_unique<TimelineManager>(port_manager_);
+    prepare_conduit_filters_();
 }
 
 CommunicatorState Communicator::get_state() const {
     assert(timeline_manager_);
     return {
         port_manager_.get_message_counts(),
-        timeline_manager_->get_state()
+        timeline_manager_->get_state(),
+        message_cache_
     };
 }
 
@@ -70,6 +114,7 @@ void Communicator::restore_state(CommunicatorState const & state) {
     assert(timeline_manager_);
     port_manager_.restore_message_counts(state.port_message_counts);
     timeline_manager_->restore_state(state.timeline_state);
+    message_cache_ = state.message_cache;
 }
 
 void Communicator::send_message(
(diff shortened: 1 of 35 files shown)


src/python/libmuscle/communicator.py
```diff
@@ -249,61 +289,84 @@ def send_message(
...
+            if not milestone_iterations:
+                break  # No milestones received, we have only messages now
+
+            # Handle most deeply nested milestone first:
+            try:
+                milestone_iteration = get_most_nested_iteration(milestone_iterations)
+            except ValueError as exc:
+                raise RuntimeError(
+                    "Internal error: received milestones for incompatible iterations "
+                    f"during F_INIT: {milestone_iterations}"
```
IrisvdWerf, line 346: You could add the following back into the string returned in the error message? “This is not supposed to happen and may be a bug in libmuscle. Please report an issue.”

src/python/libmuscle/communicator.py
```diff
@@ -336,12 +413,13 @@ def receive_s_message(self, port_name: str, slot: Optional[int] = None) -> Messa
                     raise PortClosed()
                 # TODO: handle milestone, if needed
```
IrisvdWerf: remove TODO

src/python/libmuscle/timeline_manager.py
```diff
@@ -13,6 +13,39 @@
 ExpectedActions: TypeAlias = list[tuple[str, Port, list[int]]]
 
 
+def is_subiteration(it1: IterationCount, it2: IterationCount) -> bool:
+    """Check if it1 is a subiteration of it2, or if both counts are equal.
```
IrisvdWerf, line 17: is_subiteration checks whether it1 is a subiteration of it2, including the case where both have the same count. The fact that equal counts also return true may be confusing.

src/python/libmuscle/timeline_manager.py
```diff
@@ -551,8 +578,13 @@ def check_receive(self, port: Port, slot: Optional[int] = None) -> None:
                 raise PortBlocked(port, slot, expected)
 
     def record_received_message(
```
IrisvdWerf, line 580: Since this is only called from `record_received_s_message`, why not keep the functionality there?

src/python/libmuscle/communicator_state.py
```diff
@@ -1,19 +1,24 @@
...
+from libmuscle.mpp_message import MPPMessage
+from libmuscle.timeline_manager import PortAndSlot, TimelineState
 
 
 @dataclass
 class CommunicatorState:
     port_message_counts: dict[str, list[int]]
     timeline_state: TimelineState
-    # TODO: message cache
+    message_cache: dict[PortAndSlot, MPPMessage]
```
IrisvdWerf, line 11: You could also import MPPCacheType from the communicator

maarten-ic: Yes, and no. To avoid circular imports I'd have to:

```python
from typing import TYPE_CHECKING
...

if TYPE_CHECKING:
    from communicator import MPPCacheType

...
    message_cache: "MPPCacheType"
```

Not sure if that's worth it. What do you think?

src/python/libmuscle/timeline_manager.py
maarten-ic: Good point, renamed to `is_subiteration_or_equal` now.

src/python/libmuscle/communicator.py
maarten-ic: I deliberately left that out and put "Internal error" in these consistency checks, to avoid repeating "This may be a bug".

I could add it back if we think it helps? @LourensVeen what are your thoughts?

src/python/libmuscle/timeline_manager.py
```diff
@@ -272,6 +299,30 @@ def check_send_message(
             port, slot, self._iteration
         )
 
+    def check_pre_received_iteration_counts(
```
LourensVeen: I think I complained about `check` before, arguing that `record` would be better because this updates the internal state. We changed it there, so maybe it should be changed here too?

src/python/libmuscle/timeline_manager.py
```diff
@@ -551,8 +578,13 @@ def check_receive(self, port: Port, slot: Optional[int] = None) -> None:
                 raise PortBlocked(port, slot, expected)
 
     def record_received_message(
-        self, port: Port, slot: Optional[int], iteration: IterationCount
-    ) -> None:
+        self,
+        port: Port,
+        slot: Optional[int],
+        iteration: IterationCount,
+        component_iteration: IterationCount,
```
LourensVeen: Isn't that an `instance_iteration`? Although I guess the two are the same now that we've decided that for now every instance in an instance set needs to be called the same number of times. But we're the TimelineManager for an instance, and don't know anything about other instances, so I'd still say that that's the right term.

src/python/libmuscle/communicator.py
LourensVeen: I recently saw a case where MUSCLE3 declared itself to be defective while it wasn't, so maybe we should be careful with that claim. I'm fine with `Internal error`, I think people who are inclined to file issues will do so and those who aren't won't, so the exhortation is maybe not that useful anyway. Fine to leave it like this.

src/python/libmuscle/communicator.py
```diff
@@ -249,61 +289,84 @@ def send_message(
...
-            f_init_cache: The received messages.
+            The received messages on F_INIT ports (including muscle_settings_in).
         """
         finished_iteration = self._timeline_manager.start_reuse_iteration()
-        if finished_iteration is not None:
-            self._broadcast_milestone(Milestone(finished_iteration), True)
+        milestone_iteration = finished_iteration
 
-        cache: FInitCacheType = {}
+        while True:
```
LourensVeen, line 301: Okay, yes, I think this is still complicated, but as simple as we can make it. Nice job!

src/python/libmuscle/test/test_communicator_conduit_filters.py
```diff
@@ -0,0 +1,253 @@
+from typing import Union
```
LourensVeen, line 1: This whole file is 😍. Nailed it! Nice!

src/cpp/libmuscle/port.hpp
```diff
@@ -143,7 +143,9 @@ class Port : public ::ymmsl::Port {
          */
         void set_resumed(Optional<int> slot = {});
 
+#ifndef LIBMUSCLE_MOCK_PORT_MANAGER // Enable public access for tests:
```
LourensVeen: There's already a `PRIVATE` macro used here that resolves to `private` for production and to `public` for testing. So this should not be needed, and if it is then there's something else that's gone wrong. Please investigate.

src/cpp/libmuscle/communicator.cpp
```diff
@@ -33,6 +33,45 @@ using ymmsl::Settings;
 
 namespace libmuscle { namespace _MUSCLE_IMPL_NS {
 
+namespace {
+
+/** Helper method to construct a Message from an MPPMessage.
+ * 
+ * N.B. since C++ enforces only const access to .settings and .data, we don't need to
```
LourensVeen: Actually, the settings get copied into the Message by the Message constructor, but it's true that we don't have to make one explicitly. C++ is anyway a value-based language rather than a reference-based one.

src/cpp/libmuscle/communicator.cpp
```diff
@@ -33,6 +33,45 @@ using ymmsl::Settings;
...
+    );
+    if (mpp_msg.next_timestamp.is_set())
+        message.set_next_timestamp(mpp_msg.next_timestamp.get());
+    return message;
+}
+
+
+/** Helper template method to execute code for each slot of the given port.
+ * 
+ * Expects a function with arguments (Optional<int> slot, Reference port_ref)
```
LourensVeen: That's better done with std::function, because it'll allow describing the type more precisely and is more idiomatic.

src/cpp/libmuscle/port.hpp
LourensVeen: Looked at it a bit, and it seems that `TESTING` needs to be defined to make `PRIVATE` actually `public`.

src/cpp/libmuscle/tests/test_communicator_conduit_filters.cpp
```diff
@@ -0,0 +1,279 @@
...
+    return {iteration, true};
+}
+
+IterationOrMilestone I(IterationCount iteration) {
+    return {iteration, false};
+}
+
+
+
+/** Helper method to mock MPPClient.receiev ,so it gives data with correct message
```
LourensVeen: "receiev ," -> "receive, "

LourensVeen (changes requested):
Okay, nice! This is better than the previous version, and just what I hoped could be done. Great job. I have left a few comments for small stuff to fix, then we'll merge it.

src/python/libmuscle/timeline_manager.py
IrisvdWerf: Yes, I agree.

src/python/libmuscle/timeline_manager.py
IrisvdWerf: Yes, that makes sense. It’s technically the same, but I agree that `instance_iteration` is the better term here.

src/cpp/libmuscle/port.hpp
IrisvdWerf: `build/cpp/libmuscle/tests/Makefile` already compiles every test file with `-DTESTING`, so plain `PRIVATE:` already turns private into public for all tests. I think we can remove the `#ifndef` / `#endif` wrapper?

LourensVeen (approved):
Great, let's go!

PR #412, opened 2026-09-03
Muscle manager: check timelines

maarten-ic:
This PR adds:
1. Timeline checking by the muscle manager: run `ymmsl.v0_2.resolve_timelines()` before starting the simulation to ensure that timelines are connected correctly.
2. Return the timeline of a component when it registeres with the muscle manager. We'll need this for the reducer filters in an upcoming PR.

To make this all work out, I also needed to update:
- The muscle3_tester:
  - Timeline annotations added to the ports of the tester instance.
  - If the tested component doesn't have F_INIT connections, we add a Conduit to its `muscle_settings_in` port to make these timeline annotations work.
    This was needed in order to connect to an O_F port of an instance without F_INIT ports: if we do not connect `muscle_settings_in`, the tested component would live in the root timeline and we would not be able to connect the O_F port to an S port of the tester. By connecting the `muscle_settings_in` port, the tested component lives in the correct subtimeline.
  - Patch some shutdown logic in the muscle_tester to shut down faster in case the tested component crashes or doesn't start.
- Some unit tests for registering components that would refer to a `test_instance_id` that was not part of the Configuration.

### DIFF
--- a/src/python/libmuscle/manager/topology_store.py
+++ b/src/python/libmuscle/manager/topology_store.py
@@ -1,4 +1,4 @@
-from ymmsl.v0_2 import Conduit, Configuration, Ports, Reference
+from ymmsl.v0_2 import Conduit, Configuration, Ports, Reference, Timeline
 
 from libmuscle.util import generate_indices, instance_indices
 
@@ -108,3 +108,10 @@ class TopologyStore:
                 for peer_indices in generate_indices(peer_dims[len(dims) :]):
                     peers.append(base + peer_indices)
         return peers
+
+    def get_timeline(self, instance: Reference) -> Timeline:
+        """Get the timeline for the provided instance."""
+        component = instance.without_trailing_ints()
+        timeline = self.model.components[component].timeline
+        assert timeline is not None
+        return timeline
--- a/src/python/libmuscle/pytest/implementation_tester.py
+++ b/src/python/libmuscle/pytest/implementation_tester.py
@@ -46,11 +46,11 @@ class ImplementationTester:
         self._instance._communicator.set_receive_timeout(default_timeout)
         self._instance.reuse_instance()
 
-        # We have a __settings_out__ port if the tested component doesn't have F_INIT
+        # We have a __settings_in__ port if the tested component doesn't have F_INIT
         # ports: send an empty message on it to make the timeline logic work out:
-        if "__settings_out__" in test_model.ports:
+        if "__settings_in__" in test_model.ports:
             msg = Message(float("-inf"), data=Settings())
-            self._instance.send("__settings_out__", msg)
+            self._instance.send("__settings_in__", msg)
 
     def send(
         self, port_name: str, message: Message, slot: Optional[int] = None
--- a/src/python/libmuscle/pytest/implementation_tester.py
+++ b/src/python/libmuscle/pytest/implementation_tester.py
@@ -2,7 +2,7 @@ import logging
 import os
 from typing import Optional
 
-from ymmsl.v0_2 import Configuration, Operator, Reference
+from ymmsl.v0_2 import Configuration, Operator, Reference, Settings
 
 from libmuscle import Instance, Message
 
@@ -46,6 +46,12 @@ class ImplementationTester:
         self._instance._communicator.set_receive_timeout(default_timeout)
         self._instance.reuse_instance()
 
+        # We have a __settings_out__ port if the tested component doesn't have F_INIT
+        # ports: send an empty message on it to make the timeline logic work out:
+        if "__settings_out__" in test_model.ports:
+            msg = Message(float("-inf"), data=Settings())
+            self._instance.send("__settings_out__", msg)
+
     def send(
         self, port_name: str, message: Message, slot: Optional[int] = None
     ) -> None:
@@ -53,7 +59,7 @@ class ImplementationTester:
         Send a message on the specified port.
 
         Args:
-            port_name: Name of the port to send on (without 'send_' prefix).
+            port_name: Name of the port to send on.
             message: The message to send.
             slot: Optional slot number for vector ports.
         """
@@ -70,7 +76,7 @@ class ImplementationTester:
         Receive a message from the specified port.
 
         Args:
-            port_name: Name of the port to receive from (without 'receive_' prefix).
+            port_name: Name of the port to receive from.
             slot: Optional slot number for vector ports.
             timeout: Timeout in seconds. If None, uses default_timeout.
 
--- a/src/python/libmuscle/pytest/implementation_tester.py
+++ b/src/python/libmuscle/pytest/implementation_tester.py
@@ -63,6 +63,8 @@ class ImplementationTester:
             message: The message to send.
             slot: Optional slot number for vector ports.
         """
+        if port_name == "muscle_settings_in":
+            port_name = "__settings_in__"
         self._instance.send(port_name, message, slot)
 
     def receive(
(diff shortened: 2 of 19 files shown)


src/python/libmuscle/manager/topology_store.py
```diff
@@ -108,3 +108,10 @@ def get_peer_instances(self, instance: Reference) -> list[Reference]:
                 for peer_indices in generate_indices(peer_dims[len(dims) :]):
                     peers.append(base + peer_indices)
         return peers
+
+    def get_timeline(self, instance: Reference) -> Timeline:
+        """Get the timeline for the provided instance."""
+        component = instance.without_trailing_ints()
```
IrisvdWerf: where is this without_trailing_ints() defined?

src/python/libmuscle/pytest/implementation_tester.py
```diff
@@ -46,14 +46,20 @@ def __init__(
         self._instance._communicator.set_receive_timeout(default_timeout)
         self._instance.reuse_instance()
 
+        # We have a __settings_out__ port if the tested component doesn't have F_INIT
```
IrisvdWerf: Why `settings_out` rather than `settings_in`, since it will be received on an `F_INIT` port?

src/python/muscle3/muscle_manager.py
```diff
@@ -153,6 +153,14 @@ def _manage_simulation(
         )
         sys.exit(1)
 
+    # Resolve timelines for each model
+    for model_obj in configuration.models.values():
+        try:
+            v0_2.resolve_timelines(model_obj)
```
IrisvdWerf: This is also done during the initialization of the Manager: https://github.com/multiscale/muscle3/pull/412/changes#:~:text=%23%20Ensure%20timelines%20are,.root_model()). Is it necessary to do this in both places?

src/python/libmuscle/pytest/implementation_tester.py
maarten-ic: This component will be sending a message on that port, but it's indeed not really consistent with the "port name mirroring" that we do for the other ports.

I've updated it to `__settings_in__`

src/python/muscle3/muscle_manager.py
maarten-ic: I've added it here as well, because at this point the nested models are not yet flattened: if there's an issue with the timelines then the user will get an error that's easier to understand.

LourensVeen: I guess the nice solution would be to have both a `check_timelines()` and a `resolve_timelines()` function in ymmsl, then check here only to generate errors, then flatten, then resolve the flattened model (which should then always succeed) so we can run it.

But we only have a resolve function that does both, and that works well enough, so I'm okay with leaving this as-is.

src/python/libmuscle/pytest/muscle_tester.py
```diff
@@ -82,33 +86,48 @@ def _add_tester_component(
...
+            tester_ports.append(Port(port.name, tester_operator, timeline))
+
+        if not any(
+            p.operator is Operator.F_INIT for p in implementation.ports.values()
+        ):
+            # We'll connect muscle_settings_in to make the timeline logic work
             tester_model.conduits.append(
                 Conduit(
-                    f"{tester_name}.{port_name}", f"{implementation_name}.{port_name}"
+                    f"{tester_name}.__settings_in__",
```
LourensVeen: Would it make sense to just call this `muscle_settings_in`, mirroring the name like for the other ports? That way you can also test from the test code whether overriding settings works properly.

LourensVeen (approved):
Alright, this works. One open question but no need to implement that now I think, unless it's easy and makes sense. I'll leave that to @IrisvdWerf because it's your code.

src/python/libmuscle/pytest/muscle_tester.py
IrisvdWerf: I think `muscle_settings_in` can't be used for the tester's port because port names starting with `muscle_` are reserved and rejected at runtime. Or am I misunderstanding you, @LourensVeen?

src/python/muscle3/muscle_manager.py
IrisvdWerf: I opened a PR in yMMSL-Python to split these up: https://github.com/multiscale/ymmsl-python/pull/67.

src/python/libmuscle/pytest/muscle_tester.py
LourensVeen: Ah, I hadn't thought of that. But we _are_ muscle, so we should be able to use that name. I guess that may not be so easy to implement though, and few people will use this feature in the near future I'd expect.

Maybe you could check if it's easy to do? Otherwise, let's leave it like this and come back to it if anyone actually tries to use this.

IrisvdWerf: We could give the `PortManager` a private constructor flag that lets `_ports_from_declared` accept the name `muscle_settings_in` specifically when declared as O_I, instead of rejecting every `muscle_`-prefixed name. `ImplementationTester` would pass that flag when building its own Instance,  then `get_port`, `get_message_counts`, and `restore_message_counts` need to check `self._ports` before falling back to the automatic `self._muscle_settings_in`, so the declared tester port isn't shadowed by or overwritten by the auto one. 

What do you think of this, @LourensVeen?

LourensVeen: Hmm, that feels like a bunch of extra mess just to support this one use case. I hadn't considered the shadowing either, of course the tester won't ever use its built-in `muscle_settings_in` but it's still there.

But wait, can we just leave this at `__settings_in__` and then have a one-liner in `ImplementationTester.send` that checks if the port name is `muscle_settings_in` and replaces it with `__settings_in__` if that's the case?

IrisvdWerf: This would fix the naming, but this port only exists when no `F_INIT` ports are present. We should therefore also relax the condition so that the `muscle_settings_in` conduit is created even when `F_INIT` ports exist, if we want to properly test whether overriding the settings works as expected.

IrisvdWerf: We can't do it when having other F_INIT ports, because the manager requires all F_INIT ports on a component to share one timeline, but that same timeline also drives the tester's per-round step, so muscle_settings_in  fails timeline validation.

LourensVeen: I'm not following that. The tester can deal with all kinds of ports, right? And we expect the test to send on all F_INIT ports first, then maybe do something with O_I and S ports, and finally receive on O_F ports. So then it would just be an extra F_INIT port?

Of course most of the time the user won't be thinking about `muscle_settings_in`, and won't be sending anything, causing a deadlock. So maybe it should only be added if explicitly requested when creating the tester?

LourensVeen: So, with these changes, we're dropping support for testing `muscle_settings_in`, and use it only to make things work if there are no F_INIT ports. But is that what we want?

In that case, it doesn't make much sense to add that rewrite in ImplementationTester.send either, and just leaving it as an internal and hopefully not colliding `__settings_in__` makes sense.

Well, it's in there now, so maybe just leave it as an Easter egg and add that ability to explicitly request connecting `muscle_settings_in` (and then you have to send a message on it, even if only `Settings()`) for a future PR?

IrisvdWerf: Unlike every other `F_INIT` port the tester mirrors, `muscle_settings_in` is never explicitly sent by the test. Instead, the tester's setup code produces it once, before the test loop even starts. As a result, it can't keep pace when the test drives multiple rounds through the `F_INIT` ports.

In a future PR, we could catch the `PortBlocked` error raised by `muscle_settings_in` when a round is missing it, automatically send a default empty `Settings()` override, and retry. That would make it behave like a normal `F_INIT` port without requiring test code to know about this special case.

LourensVeen: Okay, let's leave this for the future and go merge this. Thanks!

Issue #411, opened 2026-09-02
Checkpointing support when using a "last" conduit filter


maarten-ic:
The current checkpointing implementation is not compatible with "last" conduit filters and needs to be updated.

Until this issue is fixed and released, the Muscle Manager will exit with an error when launching a model with "last" conduit filters and checkpointing configured:
```text
NotImplementedError: This version of MUSCLE3 does not support checkpoints when using reducer filters. See [...].
```

### Workaround

Disable checkpointing by removing the `checkpoints:` section from the yMMSL configuration.
See https://ymmsl-python.readthedocs.io/en/stable/describing_runs.html#checkpoints

---

### Proposed solution

The checkpointing needs to be updated to use the `IterationCount` for each port's message instead of the message number:

- [ ] Update iteration count mechanism: don't reset to zero. Each number in the iteration count will then reflect the total number of messages sent in a (sub)timeline, which can be used for determining checkpoint consistency.
    - current: `[0,0]`, `[0,1]`, `[1,0]`, `[1,1]`, `[2,0]`, ...
    - proposed: `[0,0]`, `[0,1]`, `[1,2]`, `[1,3]`, `[2,4]`, ...
- [ ] Store the `IterationCount` of the last message sent / received on a Port (instead of the current `message_count`)
   - [ ] Update Port to store this (per slot)
   - [ ] Update Communicator to set iteration counts instead of incrementing num messages. Include milestones in these iteration counts
- [ ] Replace the port_message_count checkpoint metadata with iteration counts
- [ ] Update the Muscle Manager to use iteration counts for checking checkpoint consistency
- [ ] Remove message number from MPPMessage
- [ ] Update Communicator get_state / restore_state to include the `reducer_cache`
- [ ] Add test cases / integration tests for checkpointing and restoring simulations with reducer filters

PR #413, opened 2026-09-03
Add support for reducer filters

maarten-ic:
This PR adds support for reducer filters ("LAST").

Notes:
- This PR builds on top of both #410 and #412. It's best to wait with reviewing until both are merged and this branch is rebased on develop.
- LAST filters are not compatible with checkpointing at this moment, see #411.

---

**Changes in this PR:**
- Support LAST conduit filters, see below for the algorithm.
- Set MMPMessage.message_count negative and disable corresponding consistency checks when LAST filters are applied. This prevents checkpointing (see #411) and muscle_manager will raise an exception when trying to use checkpoints in a simulation with reducer conduit filters.
- Add a check in the TimelineManager that the iteration counts of F_INIT messages match the timeline of a component.

**LAST conduit filters algorithm:**
- In Communicator.prepare_conduit_filters: determine the target timeline of messages sent to a peer port with last filters applied. As an example: take component `micro` which runs in the `:macro:meso` timeline
  image
  
  We then determine the target timelines for any O_I or O_F port with LAST conduit filters:
  | # LAST filters | Target timeline<br>O_I port | <br>O_F port |
  |----------------|-----------------------------|--------------|
  | 1              | `:macro:micro`              | `:macro`     |
  | 2              | `:macro`                    | `:`          |
  | 3              | `:`                         | N/A          |

  The actual name of the timeline isn't important for the remainder of the algorithm, so we just store the "depth" of the timeline (`len(timeline)` in Python). This corresponds to the number of elements in the IterationCount for messages on that timeline. This number is stored in the `Communicator._reduced_count` dict, keyed by the destination `component.port`. I'm not too happy with the name of this variable, so feel free to update if you have a better idea!
    
  N.B. Since one outgoing port can be connected (through multicasting) with many Conduits (each potentially having different amounts of LAST filters), these "reduced_counts" are keyed by peer port (`<component>.<port>`).
- When sending a message (Communicator.send_message) to a peer port with LAST filters applied, we will store the message in the `reducer_cache` (keyed by Endpoint reference to include both the peer portname and the slot)
  - When handling milestones we check if we need to send those cached messages instead of the milestone. When multiple reducer filters are applied we may need to discard the milestone altogether. For example, if we have two LAST filters applied to the O_I port from our example above, then we discard `Milestone([0,0])`, we replace `Milestone([0])` by the last sent message and set the iteration count of that message to `[0]`. `Milestone([])` is sent as is.
  - This handling of milestone also happens in `Communicator.send_message` (`Communicator.apply_reduce_filters()` to be precise), which was much easier to implement than trying to catch it before sending the milestones in the `pre_receive()` logic.

### DIFF
--- a/integration_test/test_filters.py
+++ b/integration_test/test_filters.py
@@ -115,6 +115,24 @@ def repeat_s(filters: str) -> None:
     assert micro_received == len(EXPECTED_COUNTS)
 
 
+def postprocess() -> None:
+    instance = Instance({Operator.F_INIT: ["macro", "meso", "micro"]})
+
+    while instance.reuse_instance():
+        macro = instance.receive("macro")
+        meso = instance.receive("meso")
+        micro = instance.receive("micro")
+
+        print("Received:", macro.data, meso.data, micro.data)
+        if macro.data[1] == 0:
+            # meso and micro didn't run in this iteration, so we get an empty message
+            assert meso.data is None
+            assert micro.data is None
+        else:
+            assert meso.data[1] == macro.data[1] - 1
+            assert micro.data[1] == 1
+
+
 config = """
 ymmsl_version: v0.2
 models:
@@ -149,19 +167,27 @@ models:
         ports:
           f_init: macro meso micro
         implementation: pico
+      postprocess:
+        description: Postprocessing of final actor outputs
+        ports:
+          f_init: macro meso micro
+        implementation: postprocess
     conduits:
       macro.out:
       - meso.in
       - {filters} pico.macro
       - {filters} repeat_s.macro
+      - postprocess.macro
       meso.out:
       - micro.in
       - repeat pico.meso
       - repeat_s.meso
       - repeat repeat_s.repeated_meso
+      - last postprocess.meso
       micro.out:
       - pico.micro
       - repeat_s.micro
+      - last last postprocess.micro
 """
 
 
@@ -173,6 +199,7 @@ def test_repeater_filters(tmp_path, filters):
         "micro": ("python", micro),
         "repeat_s": ("python", repeat_s, filters),
         "pico": ("python", pico, filters),
+        "postprocess": ("python", postprocess),
     }
     run_manager_with_actors(config.format(filters=filters), tmp_path, actors)
 
@@ -186,6 +213,7 @@ def test_repeater_filters_cpp(tmp_path, filters):
         "micro": ("python", micro),
         "repeat_s": ("cpp", "conduit_filters_test", "repeat_s", filters),
         "pico": ("cpp", "conduit_filters_test", "pico", filters),
+        "postprocess": ("python", postprocess),
     }
     run_manager_with_actors(config.format(filters=filters), tmp_path, actors)
 
--- a/integration_test/test_filters.py
+++ b/integration_test/test_filters.py
@@ -129,8 +129,8 @@ def postprocess() -> None:
             assert meso.data is None
             assert micro.data is None
         else:
-            assert meso.data[1] == macro.data[1] - 1
-            assert micro.data[1] == 1
+            assert meso.data[-1] == macro.data[-1] - 1
+            assert micro.data[-1] == 1
 
 
 config = """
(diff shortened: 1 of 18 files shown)


src/python/libmuscle/communicator.py
```diff
@@ -155,6 +155,21 @@ def __init__(
         self._message_cache: MPPCacheType = {}
         """Message cache for pre-received messages."""
 
+        self._reduced_count: dict[Reference, int] = {}
```
LourensVeen, line 158: Maarten doesn't like `_reduced_count` and I agree that it's not very clear. Maybe `_target_timeline_length` is better? Or `_outgoing_timeline_length`?

src/python/libmuscle/test/test_communicator_conduit_filters.py
```diff
@@ -27,8 +33,10 @@ def repeat_filter(request):
 @pytest.fixture()
 def repeater_communicator(repeat_filter, mpp_client):
     port_manager = PortManager([], None)
+    mock_manager = MagicMock()
+    mock_manager.get_timeline.return_value = Timeline(":parent3:parent2:parent1")
```
LourensVeen, line 37: It looks like this was an implicit length-2 timeline previously, and now there's an explicit length-3 timeline and a bunch of changes to test_repeater_filters below, even though this PR doesn't touch repeaters. I don't understand why?

integration_test/test_filters.py
```diff
@@ -149,19 +167,27 @@ def repeat_s(filters: str) -> None:
         ports:
           f_init: macro meso micro
         implementation: pico
+      postprocess:
+        description: Postprocessing of final actor outputs
+        ports:
+          f_init: macro meso micro
+        implementation: postprocess
     conduits:
```
LourensVeen, line 175: It doesn't seem like we have a test anywhere with both reducers and repeaters on the same conduit. Can we add that? With the implementations having been done separately I think it would be good to check that they work together as well.

LourensVeen (changes requested):
Looks good! I have some requests, but this is close to ready.

And looking at that first comment again, `_outgoing_timeline_length` would have my preference, but feel free to suggest something even better :smile:

src/python/libmuscle/communicator.py
IrisvdWerf: I'd go with `_target_timeline_length`, it names the concept, a timeline depth, rather than the mechanism, reduction, or the call site, outgoing.

src/python/libmuscle/test/test_communicator_conduit_filters.py
IrisvdWerf: `set_peer_info` now does `self._timeline = self._manager.get_timeline()` and passes that into `TimelineManager`.  `TimelineManager.record_pre_received_iteration_counts` then asserts that received messages have the same iteration depth as `self._timeline`. 

Previously,  `manager.get_timeline()` was never called, so passing a bare MagicMock() as the manager in the test was harmless. Now, for that consistency check to hold across all three peers at once, `parent1` unfiltered, `parent2` repeated, `parent3` twice-repeated, the component's own timeline depth has to equal the deepest of the three. 

Now, every iteration count adn milestone in `test_repeater_filters` had to gain a leading index to stay depth-consistent.

IrisvdWerf: The fixture has to supply a concrete `Timeline`, and it's shared by `test_repeater_filters` and `test_repeater_filters_discard_messages`, and the latter genuinely needs 3 levels of nesting. So the mocked timeline depth is pinned at 3, forcing every iteration count and milestone in `test_repeater_filters` to gain a leading index just to stay depth-consistent.

src/python/libmuscle/communicator.py
IrisvdWerf: Let's go for `_outgoing_timeline_length`

src/python/libmuscle/test/test_communicator_conduit_filters.py
LourensVeen: Ah, okay, took a while but I see.

Honestly I don't really like the way MMPServer now stores the timeline and then lets it be retrieved later for conceptual reasons, because it's supposed to just implement the protocol and not serve as a conduit for internal communication, but I didn't realise that that design had this consequence here.

Well, it is what it is, at least I understand now, thanks!

LourensVeen (approved):
Looks good! Let's get this merged :smile:

PR #418, opened 2026-09-18
Add built-in unidirectional time bridge

LourensVeen:
This adds a new built-in time bridge actor to MUSCLE3.

It's only the documentation at the moment, describing how it will work, so that can be reviewed to make sure it makes sense before I add the implementation.


docs/source/time_bridges.rst
```diff
@@ -0,0 +1,238 @@
...
+A simple case of this is when two programs run side-by-side on the same timeline. In
+that case, they'll start at the same point in time, and then take the same timesteps, so
+that after N iterations of each, they are both at the same t_N. In that case,
+information can be sent directly from an O_I port on one program to a corresponding S
+port on the other program, and vice versa, because the two programs send and receive the
+same number of messages for any given section of simulated time.
+
+If the two programs do not run on exactly the same timeline, but one has somewhat
+smaller or larger time steps than the other or the size of the steps varies, then this
+will no longer work, because the number of messages sent by one side will not match the
```
DaanVanVugt: it is not just the number of messages, you could have equal number but still the clock skew you mention

docs/source/time_bridges.rst
```diff
@@ -0,0 +1,238 @@
...
+are waiting for each other.
+
+Time bridges exist to solve these problems, and allow connecting programs (or entire
+models) if the time lines are neither identical nor nested, but similar and possibly
+variable depending on how the simulation progresses.
+
+Connecting components using a time bridge
+-----------------------------------------
+
+The standard MUSCLE3 time bridge provides for unidirectional communication between two
```
DaanVanVugt: maybe this is the 'simplest' one but not the standard one?

docs/source/time_bridges.rst
```diff
@@ -0,0 +1,238 @@
...
+When a simulation consisting of a sender, receiver, and time bridge is started, sender
+and receiver will initialise and enter their first iteration, when they send an initial
+output on their O_I ports. The time bridge will receive the message from the receiver
+on its `clock_in` port, with the receiver's current time point `t_r_cur` and its next
+time point `t_r_next` attached.
+
+The time bridge will ignore the contents of that message, but it will use the timestamps
+to decide which input messages to send to the receiver in its next message. It will then
+receive those messages from the sender if needed and available, collate them into a
+list, and send that list to the receiver, after which it will receive the next clock
```
DaanVanVugt: this sounds like the entire list is always sent, but actually you specify below what kind of aggegration functions exist. Maybe worth a pointer

docs/source/time_bridges.rst
```diff
@@ -0,0 +1,238 @@
...
+Which data the receiver needs depends on what it does and how it works, so you will have
+to configure the time bridge to work with your specific scenario. Here are a few common
+cases, assuming that the receiver is currently at ``t_r_cur`` and is trying to step to
+``t_r_next``:
+
+- The receiver uses an explicit Euler solver. In this case, its next state will be
+  calculated based on the current state and any other inputs at ``t_r_cur``, and the
+  receiver will likely want to use the latest value before ``t_r_cur``.
+
+- The receiver uses an implicit Euler solver. Its next state will be calculated based on
```
DaanVanVugt, line 82: you can generalize the euler implicit/explicit to a theta scheme perhaps (gets you crank nicholson). But the implicit Euler coupling scheme through muscle3 confuses me anyway for now

docs/source/time_bridges.rst
```diff
@@ -0,0 +1,238 @@
...
+``recv_cur_to_next`` to ``true`` or ``false``.
+
+If ``recv_cur_to_next`` is set and ``recv_last_before_next`` equals ``1``, then the last
+message before or at ``t_r_next`` will be included only once.
+
+If ``recv_last_before_cur`` and ``recv_last_before_next`` are both ``1``, then
+messages may be received twice, once as the last message before ``t_r_next`` and then
+again on the next timestep as the last message before ``t_r_cur``.
+
+At each time step, the receiver will receive a list from the time bridge, containing
```
DaanVanVugt, line 117: But that would mean that the component gets a different interface just from being connected to the interface, because it may want to unpack the list? I think it is nicer to keep the interface (single messages) the same, because two of your typical settings are resulting in single messages being sent at all

docs/source/time_bridges.rst
LourensVeen: I'm trying to say too many things at the same time here maybe. But then you can get both clock skew and deadlocks with and without the same number of messages. Although you can't get clock skew if the time steps are the same. Okay, I'll try to phrase this better.

docs/source/time_bridges.rst
LourensVeen: There's only one now anyway, and I'm not sure which other ones could exist and be useful. I can't imagine a simpler one though, so that's a good point.

docs/source/time_bridges.rst
LourensVeen: Hmm, never heard of but I looked it up and I'll put it on the list for the future. Ideally, we'd have a description of all the different coupling schemes and how to implement them, but that's a bit too much for right now.

docs/source/time_bridges.rst
LourensVeen: Good idea! I can also imagine a helper component in between the time bridge and the receiver that takes the list and processes it down to a single value somehow. But if you just want the latest value then it's nice to be able to connect directly. Should be a configuration option I think.

docs/source/time_bridges.rst
```diff
@@ -0,0 +1,249 @@
...
+A simple case of this is when two programs run side-by-side on the same timeline. In
+that case, they'll start at the same point in time, and then take the same timesteps, so
+that after N iterations of each, they are both at the same t_N. In that case,
+information can be sent directly from an O_I port on one program to a corresponding S
+port on the other program, and vice versa, because the two programs send and receive the
+same number of messages for any given section of simulated time.
+
+If the two programs do not run on exactly the same timeline, but one has somewhat
+smaller or larger time steps than the other, then we will get clock skew, where the
+simulation time in the two components doesn't match anymore and information from the
```
IrisvdWerf, line 16: I would use “simulated time” instead of “simulation time,” as the latter could be confusing. It might be interpreted as either the time it took to run the simulation or the amount of physics time being simulated.

docs/source/time_bridges.rst
```diff
@@ -0,0 +1,249 @@
...
+Time bridges exist to solve these problems, and allow connecting programs (or entire
+models) if the time lines are neither identical nor nested, but similar and possibly
+variable depending on how the simulation progresses.
+
+Connecting components using a time bridge
+-----------------------------------------
+
+The simplest MUSCLE3 time bridge provides for unidirectional communication between two
+instances' O_I and S ports. The sending side needs to have an O_I port sending the data
+needed by the receiver, and the receiver needs the corresponding S port to receive it.
```
IrisvdWerf, line 32: You could add a figure of this coupling which is generated by the yMMSL2svg:

Image

Or does the sender not have an S port connected to the bridge actor with an O_I port?

docs/source/time_bridges.rst
IrisvdWerf: If you relate this to the [theta-scheme](https://eudml.org/doc/287863):
* **θ = 0**: explicit Euler: evaluates at t_r_cur
* **θ = 1**: implicit Euler: evaluates at t_r_next
* **θ = 0.5**: Crank–Nicolson (2nd-order accurate): evaluates at the midpoint, \(t_r_cur + \Delta t / 2\)
The goal is to always obtain the timestamp that is closest to the desired timestamp.

docs/source/time_bridges.rst
```diff
@@ -0,0 +1,249 @@
...
+  the latest value before ``t_r_next``.
+
+- The receiver integrates or averages the received values, rather than using a point
+  sample. In that case, ``t_r_cur`` to ``t_r_next`` is the next integration window, and
+  it needs to receive all messages with timestamps within that range.
+
+Note that if the time step of the receiver is smaller than that of the sender, or if the
+receiver starts its simulation before the sender starts, then there may not be any
+messages with timestamps between ``t_r_cur`` and ``t_r_next``, and the receiver may
+receive an empty list.
```
IrisvdWerf, line 93: What happens if it receives an empty list?

docs/source/time_bridges.rst
```diff
@@ -0,0 +1,249 @@
...
+The types are different here for future extensibility. For now ``recv_last_before_cur``
+and ``recv_last_before_next`` can be set to ``0`` or ``1`` only, and
+``recv_cur_to_next`` to ``true`` or ``false``.
+
+If ``recv_cur_to_next`` is set and ``recv_last_before_next`` equals ``1``, then the last
+message before or at ``t_r_next`` will be included only once.
+
+If ``recv_last_before_cur`` and ``recv_last_before_next`` are both ``1``, then
+messages may be received twice, once as the last message before ``t_r_next`` and then
+again on the next timestep as the last message before ``t_r_cur``.
```
IrisvdWerf, line 115: Why does it matter how many times the messages of a certain time stamp are included?

docs/source/time_bridges.rst
```diff
@@ -0,0 +1,249 @@
...
+|         g         |        8        |      9      |        9         |
++-------------------+-----------------+-------------+------------------+
+|         h         |        9        |             |        9         |
++-------------------+-----------------+-------------+------------------+
+|         i         |        9        |     10      |        10        |
++-------------------+-----------------+-------------+------------------+
+
+If only ``recv_last_before_cur`` or ``recv_last_before_next`` is selected, then at most
+one message will be sent to the receiver. In that case, you may want to set the setting
+``recv_list`` to ``false`` (it is ``true`` by default), to tell the time bridge to just
```
IrisvdWerf, line 162: * If `recv_last_before_cur` or `recv_last_before_next` is set, can’t we automatically set `recv_list` to `false` instead of requiring the user to specify it?
* Are there more settings next to `recv_last_before_cur`, `recv_last_before_next`, `recv_cur_to_next`, and `recv_list`?

docs/source/time_bridges.rst
```diff
@@ -0,0 +1,249 @@
...
+at which ``dt`` the solver will converge on the next try, and even if we could predict
+that in principle, that solve may depend on data that we will receive after sending this
+message and then so will the maximum achievable ``dt``.
+
+If you have such an implicit integrator connected to a time bridge as a receiver, then
+you'll want to use the value of the input at whichever ``t_cur + dt`` you're stepping
+to, and that point will change as you reduce ``dt``. So the best solution in this case
+is to send ``t_cur + dt_max`` as the value for ``next_timestamp``, and configure the
+time bridge with ``cur_to_next`` set to ``true``. You'll then receive all messages
+between ``t_cur`` and ``t_cur + dt_max``, meaning that you can always pick a value close
```
IrisvdWerf: in combination with cur_to_next is true.

